PR #561, opened 2014-10-09
init speedup

nathandunn:


nathandunn:
should not be merging into master . . .

Issue #288, opened 2014-04-19
Error in search with special characters


memartone:
I searched for R6/2 in animal models and got an error:

Not Found

The requested URL /search/R6/2 was not found on this server.

Apache/2.2.15 (CentOS) Server at monarch.monarchinitiative.org Port 80

mellybelly:
@jgrethe @ccondit  related to #358  ?

ccondit:
This isn't a limitation of SciGraph services. Seems that Ringo (or perhaps something in the JSGI standard) doesn't allow encoded `/` characters in the path (for good reason). By the time `stick` gets the request for routing the path is already changed to `search/R6/2` from `search/R6%2F2`. And I'm betting the same is true of the `.` character.

I couldn't find a Ringo configuration option to change this behavior (and it behaves the same with the latest version). Maybe @kshefchek or @kltm has a better idea of how to correctly deal with this?

There are hackish solutions:
- Change the affected characters to some unaffected UTF-8 character (from the private code page or something) and then change it back after routing.
- Use a query parameter instead of a path parameter to submit the query.

Here's a simple test case for easy debugging:

```
var stick = require('stick');
var buffer = require('ringo/buffer');
var app = exports.app = new stick.Application();
app.configure('route');
app.get('/test/:term?', function(request, term){
  return { status: 200,
           headers: {"Content-Type": "text/html"},
           body: new buffer.Buffer(term) };
});
var serve = require('ringo/httpserver');
var server = new serve.Server({app: app, port: 8080});
server.start();
```

kltm:
Ugh, a win for queries over paths here. Another hack could be just matching up to the /test part (in your example) and then doing the parsing yourself. It seems to not choke if it is ignored.

``` javascript
app.get('/test/*', function(req){
  return { status: 200,
           headers: {"Content-Type": "text/html"},
           body: new buffer.Buffer(req) };
});
```

For a more long-term solution, as always, move away from ringo/stick. Amusingly, I found out that the dependencies we use for ringo are supported in npm, so that is one step closer for #665 . A nice bridge would be getting stick working under node...

jmcmurry:
Related to https://github.com/monarch-initiative/monarch-app/issues/358

PR #772, opened 2015-06-24
Possible fix for #288

ccondit:
Possible fix based on suggestions by @kltm related to #288.


kltm:
Um, I'm not sure what the monarch process is for accepting a merge. I'd assume that this should be done by default for fixes if they pass the tests. Or, in this case, prevent the system from crashing.

kshefchek:
Is it worth setting up travis since it will run our (small and outdated) testing script against any pull requests.  Or is it too redundant having both travis and jenkins?

kltm:
I'm not too sure about all the bits of monarch, but the less places the better.

PR #775, opened 2015-06-29
fixes typos in conf/README.md. Removes install-mac.sh and refactors the mac-specific code into install.sh and installRingo.sh. adds installRingo.sh which isolates platform-specific ringojs installation in one file. install.sh now also calls 'gulp assemble' to generate the conf/monarch-team.json from the .yaml file. Deletes /stick directory which is already being installed via install.sh. Deletes bogus 'esc' file that looks like it was erroneously added (the 'esc' file contains 'git diff' output, which is very confusing when you see it referred to in a real 'git diff'. Fixes update_dependencies.sh to make clear which version of phenogrid is being used. Updates package.json to include gulp.

DoctorBud:
- fixes typos in conf/README.md.
- Removes install-mac.sh and refactors the mac-specific code into install.sh and installRingo.sh.
- adds installRingo.sh which isolates platform-specific ringojs installation in one file.
- install.sh now also calls 'gulp assemble' to generate the conf/monarch-team.json from the .yaml file.
- Deletes /stick directory which is already being installed via install.sh.
- Deletes bogus 'esc' file that looks like it was erroneously added (the 'esc' file contains 'git diff' output, which is very confusing when you see it referred to in a real 'git diff'.
- Fixes update_dependencies.sh to make clear which version of phenogrid is being used.
- Updates package.json to include gulp.


kshefchek:
Merged, let's add a test to start-server.sh to check to see if ringo has been installed and if not install it automatically.

Issue #779, opened 2015-07-02
Installation error on CentOS 6.6


yuanzhou:
Anyone know how to resolve this?

![capture](https://cloud.githubusercontent.com/assets/195873/8484600/83b56c5a-20c6-11e5-90c7-141e55e761cd.PNG)

PR #780, opened 2015-07-02
monarch-app 779 - Fixes typo that prevents server startup.

DoctorBud:
Fixes https://github.com/monarch-initiative/monarch-app/issues/779


kshefchek:
Oops, we both fixed it.

PR #804, opened 2015-07-09
Fixes a bug in the dovechart code. The bug is that if you descend into a category by clicking the disclosure triangle, adding a 'breadcrumb', and then ascend by clicking on a higher level breadcrumb, the app will appear to be working. But if you try to descend again, you will get a Javascript error. Fix is simple... Don't over-delete the breadcrumbs on ascent.

DoctorBud:
Fixes a bug in the dovechart code. The bug is that if you descend into a category by clicking the disclosure triangle, adding a 'breadcrumb', and then ascend by clicking on a higher level breadcrumb, the app will appear to be working. But if you try to descend again, you will get a Javascript error. Fix is simple... Don't over-delete the breadcrumbs on ascent.

### DIFF
--- a/widgets/dove/js/dovechart.js
+++ b/widgets/dove/js/dovechart.js
@@ -46,7 +46,7 @@ monarch.dovechart = function(config, tree, html_div){
     self.parents.push(tree.getRootID());
     self.html_div = html_div;
     self.tree = tree;
-    
+
     self.tooltip = d3.select(html_div)
         .append("div")
         .attr("class", "tip");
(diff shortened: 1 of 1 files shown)


DoctorBud:
There may be an GitHub issue for this bug, but I did not find it in when I looked.

kshefchek:
Thanks @DoctorBud! We definitely need to add unit tests here.  The code for this widget is a bit of a mess (in the middle of refactoring) but would love to hear your feedback.

PR #812, opened 2015-07-13
Fixes package.json to point to the npmify-phenogrid branch for the phenogrid npm module. Uses /node_modules/phenogrid instead of /widgets/phenogrid Deletes /widgets/phenogrid Fixes 'Name:' to be 'name:' in monarch-team.yaml. Uses async:true for Ajax calls to phenotype_list.json, to address a Safari issue where the jquery timeout is ignored (see http://stackoverflow.com/questions/31140212/failed-to-load-resource-request-timed-out-on-safari and http://propercode.com/wordpress/?p=32). Uses non-jQuery client mode (Phenogrid.createPhenogridForElement) to load and initialize phenogrid. Deletes update_dependencies.sh Adds a HACK so that if serverUrl is specified, the assumption is that serverUrl points to a /widgets/phenogrid server rather than a /node_modules/phenogrid server.

DoctorBud:
Fixes package.json to point to the npmify-phenogrid branch for the phenogrid npm module.
Uses /node_modules/phenogrid instead of /widgets/phenogrid
Deletes /widgets/phenogrid
Fixes 'Name:' to be 'name:' in monarch-team.yaml.
Uses async:true for Ajax calls to phenotype_list.json, to address a
Safari issue where the jquery timeout is ignored (see
http://stackoverflow.com/questions/31140212/failed-to-load-resource-request-timed-out-on-safari
and http://propercode.com/wordpress/?p=32).
Uses non-jQuery client mode (Phenogrid.createPhenogridForElement) to
load and initialize phenogrid.
Deletes update_dependencies.sh
Adds a HACK so that if serverUrl is specified, the assumption is that
serverUrl points to a /widgets/phenogrid server rather than a
/node_modules/phenogrid server.


kshefchek:
@harryhoch and @yuanzhou is this okay to merge?

PR #911, opened 2015-08-09
- Adds new step type to behave that allows for waiting on a DOM id. - Adds a maui_phenogrid test. - Moves implicitly_wait to environment.py, where it establishes a wait time for the session. - Note that the implicitly_wait doesn't seem to work in all cases, which is why the explicit wait step was added in the 'slow page' step. - Adds set_window_size to the webdriver instance so that PhantomJS can 'see' the autocomplete field.

DoctorBud:
- Adds new step type to behave that allows for waiting on a DOM id.
- Adds a maui_phenogrid test.
- Moves implicitly_wait to environment.py, where it establishes a wait time for the session.
- Note that the implicitly_wait doesn't seem to work in all cases, which is why the explicit wait step was added in the 'slow page' step.
- Adds set_window_size to the webdriver instance so that PhantomJS can 'see' the autocomplete field.

Verified via TravisCI and via PhantomJS on MacOSX.

### DIFF
--- a/tests/behave/environment.py
+++ b/tests/behave/environment.py
@@ -6,6 +6,7 @@
 ####
 
 import os
+import time
 from selenium import webdriver
 
 ###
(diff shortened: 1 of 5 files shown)


DoctorBud:
Hmmm. The tests work fine on Firefox Selenium on my Mac. Not so good on the TravisCI. The basic problem is that the slow-loading pages aren't loaded by the time that behave gets around to checking for certain strings in the HTML text. I'll continue to debug this by adding print statements and figuring out what Selenium is 'seeing' and when.

PR #940, opened 2015-09-01
Fixes the use of NodeJS request to be compatible with the ringo/request.  Mostly this involved the use of the 'useQuerystring' option to ensure that parameters are encoded in form-url-encoded mode.  Removes use of getTemplate() in webapp and the explicit use of mustache in favor of calling puptent's render() method.  Deletes lots of dead code.

DoctorBud:
Fixes the use of NodeJS request to be compatible with the ringo/request.

Mostly this involved the use of the 'useQuerystring' option to ensure
that parameters are encoded in form-url-encoded mode.

Removes use of getTemplate() in webapp and the explicit use of mustache
in favor of calling puptent's render() method.

Deletes lots of dead code.

### DIFF
--- a/lib/monarch/api.js
+++ b/lib/monarch/api.js
@@ -1395,7 +1395,7 @@ bbop.monarch.Engine.prototype.getDiseaseAssocsFromOrthologs = function(gene) {
         if (gene.orthologs[i].organism == 'Homo sapiens'){
             var disObj = engine.fetchDiseaseGeneAssociations(gene.orthologs[i].ortholog.id,'gene');
             if (disObj){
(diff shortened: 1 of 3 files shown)


kshefchek:
@DoctorBud I'm having issues rendering the solr results as tables with your PR, see {hostname}/gene/NCBIGene:6469 and click on one of the tabs (other than compare or overview) as an example.

cmungall:
@kshefchek will you file a ticket?

kshefchek:
@cmungall I reverted the PR and discussed with @DoctorBud so it is a non-active issue, should I still file a ticket?

cmungall:
no

PR #972, opened 2015-09-29
Updates followed by Phenogrid 1.1.0

yuanzhou:
1. removed the use of `htmlPath` and `imagePath`, phenogrid no longer uses external image files as well as separate html files for popup dialogs
2. updated phenogrid version number to 1.1.0 in package.json to sync with the phenogrid npm updates

@kshefchek @DoctorBud Please review and merge.

### DIFF
--- a/js/Analyze.js
+++ b/js/Analyze.js
@@ -728,16 +728,6 @@ function AnalyzeInit(uploaded_data){
         
 
         var phenotypes  = text.split(/[\s,]+/);
-        // Old way - Zhou
-	/*
-	$("#phen_vis").phenogrid({
-				phenotypeData: phenotypes,
-				targetSpeciesName: species,
-				owlSimFunction: urlParams.mode,
-				geneList: urlParams.geneList,
-				providedData: urlParams.user_input
-			});
-	*/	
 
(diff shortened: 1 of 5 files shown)


yuanzhou:
@nathandunn it seems something is wrong with Github. I saw this pull request was tagged as in progress before I submitted the pull request.

nathandunn:
@yuanzhou   Not sure why / how the "in progress" label got added.   I have been out of the loop on this for awhile.

yuanzhou:
@nathandunn maybe try to delete the forked monarch-app repo from your github account? I remembered similar situation happened couple months ago between us.

yuanzhou:
@kshefchek do you have any idea of why the Travis build failed? I only made minor changes which I don't think will cause the build failure. The reason that I sent this PR is because I believe there are many changes in monarch-app now, and I don't want to cause any confusions.

kshefchek:
A few tests are dependent on the latest scigraph load(which we are currently reloading).  So the switch to the production scigraph service is the culprit.

yuanzhou:
Thanks @kshefchek

PR #1015, opened 2015-10-15
Ready to Merge, but waiting on some scheduling issues... Migrates modules to npm modules, including jquery, bootstrap, d3 and more.

DoctorBud:
Run with `npm run start` or `./start-server.sh`

The recently completed Ringo-to-NodeJS refactor was confined to the issues of server behavior and how the server delivers content to the browser.

In contrast, this very large PR is about refactoring (as minimally as possible) the front-end behavior and asset subsystem. One goal behind this PR is to use NPM-installed most-recent versions of software modules that are currently stored within the git repo and are wildly out of date. This will ease maintenance and understanding of the codebase, as well as enabling more rapid development. Along the way towards making this happen, I ended up fixing a lot of risky front-end code
- Without loss of functionality (based upon my eyeballs and tests), I was able to eliminate the monarch-app version of pup-tent in favor of the NPM-available version.
- Similarly, I replaced the hardcoded d3.js with an NPM version.
- I deleted raphael.js, which was apparently not used.

I got ambitious and eliminated the old stale 3.0.0 bootstrap.min.js in favor of the current 3.3.5 version. Unfortunately, this produced some undesirable UI effects that I have been working through. The usual cause is that we had CSS workarounds for 3.0.0 behaviors and these
workarounds are either no longer necessary or are inappropriate for 3.3.5.

Some of the packages moved to node_modules include pup-tent, bootstrap,
d3, jQuery and others.

RequireJS is now used for asynchronous module loading and dependency
management. When using the webpack bundler, then RequireJS is not used
at all.

Fixes more bootstrap issues (tabs look better, analyze page looks
better). Makes useBundle controllable via an ENV var.

Improves bundlability of the JS and CSS

More bootstrap UI cleanup. Moves more CSS into bundle

RequireJS for non-bundled mode and modern jquery-ui. delete old jquery/jquery-ui
## 

Other changes:
- When in non-production mode, an error page produces a visible stack trace and
  error message, instead of hiding it in the HTML comment.
- Improves the useBundle behavior
- Fixes loader() so that it occurs during jquery(document).ready()
- Adds tables.js back in
- Changes loader() to use jquery.ready() for both bundle and non-bundle
- Adjusts Analyze.js to not use window.onload for invoking Phenogrid, since the prior loader() change takes care of this deferral.
- Fixes most of the behave tests
- Adds explicit Wait to the autocomplete test step
  all tests pass, except for a slow-running one that I disabled
  checkpoint
- Updates NPM packages
- Downgraded HapiJS to 10.5 due to CORS issues that I will get to when 11.0 stabilizes and has better docs.
- Lots of fixes to get the 3 versions (legacy, devserver, production) running the same code and CSS
- Fixes the way that the bundled /dist is handled so that pup-tent is not used for bundled resources. This is due to known (and easily fixable) bug in pup-tent; basically, it fails with NodeJS (and possibly RingoJS) when trying to deliver binary files like PNG. Rather than fix it, I'm continuing the workaround.
- Increases delays for selenium Waits due to slow servers over the weekend.
- All 3 NodeJS versions pass their behave unit tests
- start-server.sh, npm run dev, and npm run build/bstart all work.
- Fixes Feedback menu trigger to look nice and have a pencil icon.
- Adds yuicompressor as an NPM dev dependency.
- Updates scripts/release-js.pl to use the installed yuicompressor.
- Migrates fixes that were erroroneously placed in the generated monarch.js into the appropriate source files.
- Adds loaderHeader/Footer JS fragments to sandwich the components of the generated monarch.js


kshefchek:
I'm currently investigating why our tests are failing, will rerun this when it's fixed.

cmungall:
Testing now
- ontology hierarchy view appears broken
- Navigation between tabs appears broken on phenotype pages (but not disease or gene)

kshefchek:
@DoctorBud our behave tests are now fixed, looks like the PR just fails this one:
https://github.com/monarch-initiative/monarch-app/blob/master/tests/behave/maui_gene.feature#L7

kshefchek:
@DoctorBud how far away are we from merging this PR?  Some of these things could be immediately useful, specifically passing server variables to the client.  This would be useful for our autocomplete widget where I've temporarily hardcoded the SciGraph server url (I was going to fix, but this would be a lot better).
Also, monarch.js is automatically generated from merging the files in /js/lib/*.  I see that you've edited it directly.  What would be the best approach in terms of bundling these files?  Could we do this with gulp?

DoctorBud:
@kshefchek and I had a good discussion today. I had not known about monarch.js being a derived file (via release_js.pl), so I've been making changes to it and not running 'make' or 'release_js.pl'. Now that I know the correct flow, I can apply my changes to the 4 component files and automate the packing of these files as part of start-server.sh (npm run start).

Even easier would be for us to declare monarch.js to be the new source file and to delete its components and the concatenation part of release_js.pl.

kshefchek:
+1

cmungall:
After running `npm install` successfully:

```
./start-server.sh 
Starting server. Platform: NodeJS. Environment: dev. Port: 8080.
defaultConfigFile: conf/server_config_dev.json golrConfigFile: conf/golr-conf.json

/Users/cjm/repos/DoctorBud/monarch-app/node_modules/hapi/lib/index.js:5
const Server = require('./server');
^^^^^
SyntaxError: Use of const in strict mode.
    at Module._compile (module.js:439:25)
    at Object.Module._extensions..js (module.js:474:10)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:312:12)
    at Module.require (module.js:364:17)
    at require (module.js:380:17)
    at Object.<anonymous> (/Users/cjm/repos/DoctorBud/monarch-app/lib/monarch/web/webapp.js:247:16)
    at Module._compile (module.js:456:26)
    at Object.Module._extensions..js (module.js:474:10)
    at Module.load (module.js:356:32)
```

DoctorBud:
Sorry Chris, I forgot to update the README to Node V4.0. Here's a good version list:

npm: '3.4.0',
node: '4.2.2',

I'll update the README. I avoided going to NodeJS 4.0 until the last possible minute, but it seems to be where the Node World is nowadays, so recently I committed to 4.0 and forgot to update the README.

I will have a fix to the doc soon.

kshefchek:
@DoctorBud we will have to install those versions of npm and node on our servers.  If this takes a while, how hard is it to switch back to the older versions?

kshefchek:
I was able to get these versions of npm and node on our beta server; however, I'm still seeing some build failures related to node-gyp.  @DoctorBud do you know what to make of the logs? (rename the file with a .tar.bz2 extension)
[MON-BETA-JOB1-223.txt](https://github.com/monarch-initiative/monarch-app/files/37549/MON-BETA-JOB1-223.txt)

DoctorBud:
I'm working on a regress-node-to-0-12-2 branch which should allow us to work on the UCSD nodes without installing Node 4.4.2. Node 4.4.2 is problematic on the UCSD CentOS nodes because we may not have (has this been verified yet?) the sudo privilege necessary.

The future for NodeJS is now, however, and we really should plan on getting back on the 4.x version. I found some instructions that look credible here:

[Install GCC 4.8.1 and other Scientitic Packages via Yum on CentOS](http://linuxtoolkit.blogspot.ca/2014/11/install-gcc-481-and-other-scientitic.html)
[Install g++/gcc 4.8.2 in CentOS 6.6](http://qiita.com/giwa/items/28c754d8fc2936c0f6d2)
[How to compile gcc-4.8.2 on CentOS 6.6](http://wiki.mikejung.biz/Gcc_CentOS)

PR #1031, opened 2015-10-23
Simsearch -  get type from scigraph if null

kshefchek:
See https://github.com/monarch-initiative/phenogrid/issues/210, this is a pretty small change, @nlwashington or @cmungall could you review and +1?


yuanzhou:
@kshefchek Thanks for addressing this issue. Does this mean all the simsearch and compare returned targets will have a type?

kshefchek:
@yuanzhou yes.

PR #1046, opened 2015-10-29
allow trailing slash on /score route. Fixes #1044

### LINKED ISSUE #1044: Ensure routes with a trailing slash still work
The examples here don't work:

http://beta.monarchinitiative.org/page/services

When we switched from ringo to nodejs, it got more particular about routes

So this fails:
/score/?annotation_profile={%22id%22%20:%20%22MONARCH:Example2%22,%22features%22%20:%20[{%22id%22:%22HP:0000505%22},{%22id%22:%22HP:0000479%22},{%22id%22:%22HP:0001010%22},{%22id%22:%22HP:0000044%22}]}

The solution is to remove the trailing slash:

/score?annotation_profile={%22id%22%20:%20%22MONARCH:Example2%22,%22features%22%20:%20[{%22id%22:%22HP:0000505%22},{%22id%22:%22HP:0000479%22},{%22id%22:%22HP:0001010%22},{%22id%22:%22HP:0000044%22}]}

We should either make the routes more liberal OR at least update our docs so that we use the correct form

cmungall:
@DoctorBud look ok?

### DIFF
--- a/lib/monarch/web/webapp.js
+++ b/lib/monarch/web/webapp.js
@@ -2628,6 +2628,10 @@ function scoreHandler(request) {
 web.wrapRouteGet(app, '/score', '/score', [], scoreHandler, errorResponse);
 web.wrapRoutePost(app, '/score', '/score', [], scoreHandler, errorResponse);
 
+// Also allow trailing slashes
+web.wrapRouteGet(app, '/score/', '/score/', [], scoreHandler, errorResponse);
+web.wrapRoutePost(app, '/score/', '/score/', [], scoreHandler, errorResponse);
+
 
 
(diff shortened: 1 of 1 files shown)


kshefchek:
+1

DoctorBud:
Looks good to me. It may cause problems with RingoJS, which treats foo/ and foo identically. But we can cross that bridge if we ever need to go back to Ringo. If we find that adding the trailing slash is common enough, then I'd advocate enhancing web.wrapRouteXXX() with a parameter addTrailingSlash=true that would add the aliased routes for Node only.

cmungall:
I consider the bridge to ringo well and truly burned...

PR #1050, opened 2015-10-30
Add collaborations and partners page

kshefchek:
This is a preemptive PR for making the collaborations and partners page.  See https://github.com/monarch-initiative/impact/issues/24


kshefchek:
@jmcmurry I still need to make a few updates such as adding new logos and updating the reeldx logo, but in general I'm ready for your feedback.

kshefchek:
collab-page
collab-page-2

jmcmurry:
Thanks @kshefchek looks good; please swap out the illegible omim logo for the one I just hacked up. Beware that it is a png and not a jpg, so you will need to change the file path... also the LBNL logo is missing it seems. Also, if it is possible to make the logos more similarly sized, that would be great. Don't knock yourself out though. The lack of consistency in aspect ratios makes this a challenge. @mellybelly may have additional thoughts.

jmcmurry:
We also need to be clearer about what that spectrum refers to as it doesn't include any sources. I think that for the website, it may be worthwhile to just have everybody in there; sources, consumers, and supporters. @mellybelly ?

jmcmurry:
@mellybelly can this be merged or do we need to get authorization? Not sure where things stand at the moment.

PR #1055, opened 2015-11-03
Split "slow page" behave test into two steps

kshefchek:
We sometimes have issues with the test 

```
    I go to slow page "{page}" and wait for id "{id}"
```

Where context.browser.get(context.target + page) is called and while the page is loading

```
    element = WebDriverWait(context.browser, 200).until(EC.presence_of_element_located((By.ID, id)))
```

is executed and presence_of_element_located is satisfied by an element loaded from a previous test.  Note that this only seems to be an issue on travis.

This fix splits the test into two independent tests:

```
    I go to page {page}
    then wait for id {id}
```

@doctorbud does this fix make sense?

### DIFF
--- a/tests/behave/maui_gene.feature
+++ b/tests/behave/maui_gene.feature
@@ -5,6 +5,7 @@ Feature: Gene page description
 
  @ui
  Scenario: text based description of the gene appears in a user-friendly way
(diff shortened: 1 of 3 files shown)


kshefchek:
This doesn't fix it, see https://travis-ci.org/monarch-initiative/monarch-app/builds/89041556#L1843

PR #1058, opened 2015-11-05
REVIEW: Adds CONTRIBUTING.MD. Minor update to README.md

DoctorBud:
This is the first draft of CONTRIBUTING.md, which will provide guidance to existing and new Monarch developers and contributors with regard to our GitHub Flow-style process.

Please review this document and add comments to this Pull Request. I will iterate for a few days until the feedback stabilizes into someone saying, 'ship it' or +1.

Thanks,
Dan


harryhoch:
@DoctorBud, thanks for doing this.

Should we add something about deleting branches once PRs are accepted?

cmungall:
+1 to merging docs.

Also, should we have some docs for syncing your branch (what you have there is syncing master before making a branch)? I'm happy to just advocate the git rebase approach here.

DoctorBud:
I will add something about deleting branches. I tend not to do it, since all of the pollution is local to my fork. But I know that I could do it if I wanted. A friend of mine recently shared this on Facebook ... https://kparal.wordpress.com/2011/04/15/git-tip-of-the-day-pruning-stale-remote-tracking-branches/

Regarding the syncing your branch question from Chris. You are correct, I described how to sync master before creating a new branch. What you are talking about is the idea of periodically 'merging' or 'rebasing' from the latest fetched upstream master. That definitely needs to be added to the document. I'm going to advocate rebase over merge when performing this operation, since it leaves behind a cleaner commit history.

So two things I gotta document:
- Deleting old branches
- Refreshing in-development branches with upstream changes to ensure no merge conflicts.

Thanks for the feedback.

harryhoch:
@DoctorBud , fair point. I retract. If people are only branching in local repos, it's up to them.

that said, we should clean up monarch-app some times soon. lots o' crud there.

DoctorBud:
@kshefchek The 'Fork' part happens only once, at which point you repeat the Branch->PR->Merge cycle. So basically, they are talking about the same thing. I tried to clarify this in my document by making clear that the 'Fork-and-Clone' step happens once per developer, after which they do the Branch->PR->Merge over and over again. Does that help?

I'll double-check the links to make sure they aren't confusing things and that they correspond to what I wrote above.

@kshefchek you are correct in that the first link blurs the distinction between the one-time Fork and the many-time Branch->PR->Merge. I will add a note when referring to the doc.

kshefchek:
I'm completely open to the workflow, and it seems almost the same to me.  For the sake of argument, this work flow adds a tiny bit of complexity:
- You have to manage two remotes - origin, and upstream, although theoretically you should only be pulling from upstream, or alternatively resetting per the example in your contributing doc
- Branches are spread across different repositories, it might slightly harder to navigate versus having everything under the same repository

I think these are relatively small negatives and easy to adapt to once adopted.

The advantages
- Users without permissions have a way to contribute via this method
- In a sense it keeps the main repository "clean" from experimental branches, and any clutter generated from people making branches and leaving them behind

I think it could make sense for monarch-app as it grows.  Maybe not for something like phenogrid or dipper where typically 1-2 developers work on the project.  But that's just my 2 cents.

DoctorBud:
I think that I'm at a point of diminishing returns with the CONTRIBUTIONS doc, and that we should merge it and crowdsource (or at least, teamsource) the continued evolution of the doc. Once we get a few people trying to use the doc and the workflow, we'll get more useful feedback and can improve the document in a more targeted way.

harryhoch:
+1

Issue #1032, opened 2015-10-23
Missing properties in compare API returned JSON


yuanzhou:
@kshefchek  I think this issue is related to #1024  and #1004 

http://beta.monarchinitiative.org/compare/HP:0000726+HP:0000746+HP:0001300+HP:0002367+HP:0000012+HP:0000716+HP:0000739+HP:0001332+HP:0001347+HP:0002063+HP:0002067+HP:0002172+HP:0002322+HP:0007159+HP:0009466+HP:0001972+HP:0003502+HP:0005190+HP:0001763+HP:0003318+HP:0001371+HP:0001380+HP:0001394+HP:0001156+HP:0001159+HP:0001004+HP:0010886+HP:0006530+HP:0004792+HP:0002816+HP:0000954+HP:0000974+HP:0000765+HP:0000602+HP:0002240+HP:0030084+HP:0000049+HP:0000028+HP:0000175+HP:0000204+HP:0000202+HP:0000316+HP:0000343+HP:0000349+HP:0002023+HP:0000327+HP:0002055+HP:0000394+HP:0000463+HP:0000431+HP:0000494+HP:0000484+HP:0000486+HP:0000508+HP:0009748+HP:0007018+HP:0001773+HP:0001769+HP:0003311+HP:0001544+HP:0001508+HP:0003196+HP:0001249+HP:0001187+HP:0001169+HP:0012774+HP:0002650/ZFIN:ZDB-GENO-080606-613,ZFIN:ZDB-GENO-991201-127,ZFIN:ZDB-GENO-991130-127,ZFIN:ZDB-GENO-101216-3

This compare only returns the `a` part in the JSON, missing `b`, `metadata` and `resource`.

kshefchek:
Also related: https://github.com/monarch-initiative/monarch-app/issues/1005

yuanzhou:
`metadata` issue has been addressed in PR #1069 , my question is even if there's no matches `b`, why we still have `metadata.maxMaxIC` (used to generate color scale for Uniqueness calculation method in Phenogrid)? Does this mean the `metadata.maxMaxIC` is a fixed value? If so, we don't even need to get the `maxMaxIC` from `metadata` from each owlsim result. @harryhoch any thoughts?

Example: 

[...]

Issue #1005, opened 2015-10-14
Sometimes `metadata` property is missing in the compare results


yuanzhou:
I'm getting a list of genotypes that are associated to a specific gene, then use the `compare` api to get matches between the genotypes and a list of phenotypes. I tried two mouse genes, during the compare phrase, the returned matches doesn't have the `metadata` property for the second genotypes/phenotypes query. 

![capture](https://cloud.githubusercontent.com/assets/195873/10488202/99acd99e-7265-11e5-8935-68741646318b.PNG)

Please click the image to see the full version.

kshefchek:
Any chance you could the paste the failing queries here?  What is expected in the metadata field?

yuanzhou:
Here is a query that returns the `metadata`:

http://beta.monarchinitiative.org/compare/HP:0000726+HP:0000746+HP:0001300+HP:0002367+HP:0000012+HP:0000716+HP:0000739+HP:0001332+HP:0001347+HP:0002063+HP:0002067+HP:0002172+HP:0002322+HP:0007159+HP:0009466+HP:0001972+HP:0003502+HP:0005190+HP:0001763+HP:0003318+HP:0001371+HP:0001380+HP:0001394+HP:0001156+HP:0001159+HP:0001004+HP:0010886+HP:0006530+HP:0004792+HP:0002816+HP:0000954+HP:0000974+HP:0000765+HP:0000602+HP:0002240+HP:0030084+HP:0000049+HP:0000028+HP:0000175+HP:0000204+HP:0000202+HP:0000316+HP:0000343+HP:0000349+HP:0002023+HP:0000327+HP:0002055+HP:0000394+HP:0000463+HP:0000431+HP:0000494+HP:0000484+HP:0000486+HP:0000508+HP:0009748+HP:0007018+HP:0001773+HP:0001769+HP:0003311+HP:0001544+HP:0001508+HP:0003196+HP:0001249+HP:0001187+HP:0001169+HP:0012774+HP:0002650/MGI:3811489,MGI:2170676,MGI:5443937,MGI:3628465

And this one doesn't:

[...]

PR #1066, opened 2015-11-10
Add metadata to simsearch and compare json when no results are found

kshefchek:
See issues #1032 and #1005, this PR adds the metadata field to simsearch and compare results that return no results.

### DIFF
--- a/lib/monarch/api.js
+++ b/lib/monarch/api.js
@@ -4229,8 +4229,15 @@ bbop.monarch.Engine.prototype.makeSimComparisonResults = function(stuff, metric,
     if (stuff.length > 0) {
         system_stats = stuff[0].system_stats;
     }
-    //console.log(JSON.stringify(stuff));
     
+    //HARDCODE ALERT
+    //TODO this should be the result of an owlsim call
+    if (system_stats != null && typeof system_stats != 'undefined') {
+        system_stats.metric_stats = {metric : metric, maxscore : '100', avgscore : '60', stdevscore : '4.32', comment:'These stats are approximations for this release'};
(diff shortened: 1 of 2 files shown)


yuanzhou:
Thanks for addressing this metadata issue. I think metadata is needed only when we have the "b" in the JSON since it contains all the matches. It's fine without metadata if there's no "b" since we use the matches in "b" to render the grid cells and metadata for cell colors.

In phenogrid, can I assume that if there's no "b" it means no matches found?

yuanzhou:
I'll remove the metadata check from phenogrid, only check to see if "b" is present then display no data message in its absence.

yuanzhou:
It seems this PR doesn't fully address the `metadata` issue. I tested at : 

http://localhost:8484/analyze/phenotypes/?input_items=HP:0012594+MP:0000433&mode=compare&gene_list=NCBIGene:100856649,NCBIGene:616699,NCBIGene:257634,NCBIGene:791744,ZFIN:ZDB-GENE-980526-116,MGI:99783,ZFIN:ZDB-GENE-980526-347,ZFIN:ZDB-GENE-980526-116,MGI:99783,ZFIN:ZDB-GENE-980526-347

and the grid cell color is incorrect. I looked into the code: https://github.com/monarch-initiative/phenogrid/blob/master/js/dataloader.js#L215

and then noticed that we still need `metadata.maxMaxIC` to generate the correct color scale. 

c976ad3a-726a-11e5-9d34-180a359f7f2b

This PR added `metadata.metric_stats` property but the `maxMaxIC` is still missing.

PR #1074, opened 2015-11-14
Merge master with stage branch

kshefchek:


kshefchek:
The "phenotype pubmed browser" button on the homepage goes to the analyze phenotypes page, is this correct?

PR #1092, opened 2015-11-19
- Fixes pup-tent usages so that text-ish media like CSS files are delivered with UTF8 encoding, while binary such as PNG is delivered with binary encoding. - Fixes layout issues with tabs on specific disease/gene/... pages. - Fixes phenogrid FontAwesome bug that caused the info icon to be a letter Z.

DoctorBud:
- Fixes pup-tent usages so that text-ish media like CSS files are delivered with UTF8 encoding, while binary such as PNG is delivered with binary encoding.
- Fixes layout issues with tabs on specific disease/gene/... pages.
- Fixes phenogrid FontAwesome bug that caused the info icon to be a letter Z.
- Fixes duplicate InitPhenogrid() call in phenogrid-loader-nospecies, which caused duplicate species to appear.
- Fixed variant page to have proper container heirarchy.


kshefchek:
+1

PR #1098, opened 2015-11-22
Merge master into production

kshefchek:
This branch is the result from a hard reset of the production branch to the current master branch.  So if any changes were made to the production branch, and not master, they will be lost.


cmungall:
+1

PR #1103, opened 2015-11-25
README has better NPM install instructions. Autocomplete menu consistency. CSS cleanup.

DoctorBud:
- Updates README with nvm and n instructions and better install instructions.
- Fixes monarch-common.css to override jquery-ui styles in the same way that phenogrid's jquery-ui-modified.css does, but to use explicit selectors to reduce non-deterministic CSS matching.
- This fixes the autocomplete menus throughout monarch to be consistent with each other.
- Deletes monarch-main.css.
- Adds route aliases for slash-terminated landing pages like /disease and /disease/. Same with /simsearch/phenotype.


kshefchek:
+1

PR #1112, opened 2015-12-02
Updates production servers

kshefchek:
Changes production scigraph and golr servers to use our staging servers.


cmungall:
I'm going to merge this, it won't actually take effect until we merge master into production and redeploy

PR #1129, opened 2015-12-18
Phenogrid version update and behave test updates

yuanzhou:
1. updated phenogrid version to v1.2.3 in package.json
2. updated corresponding behave tests since this new version of Phenogrid supports multiple instances, which caused ID selector changes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "jquery": "^2.1.4",
     "jquery-ui": "^1.10.5",
     "mustache": "^2.2.0",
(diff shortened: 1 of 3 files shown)


yuanzhou:
Sending a pull request and wait for the TravisCI build is a good way to discover changes that I've missed.

kshefchek:
+1

yuanzhou:
@kshefchek thanks for merging this PR. I think Dan will address the fonts path issue and I'll look into it too.

PR #1135, opened 2015-12-22
Fixes issues with phenogrid and fontawesome. Fixes many JS issues reported by ESLint. Reduces travis delay after server start. Changes default browser for browsersync.

DoctorBud:
- Fixes issues with phenogrid and fontawesome.
- Enables eslint, but disables it for now. Fixes many JS issues reported by ESLint. See the LINT variables in webpack.make.js if you want to turn it on. Or use Sublime's SublimeLinter, which will honor the `.eslintrc` file.
- Fixes issue with BrowserSync where Firefox would reload a page instead of following a link.
- Changes the default browser for BrowserSync to Firefox, which works better for unix systems.
- Removes USE_LOG_FETCH=0 from `npm run dev` to enable easier overriding (e.g., `USE_LOG_FETCH=1 npm run dev`
- Modifies `install.sh` to use the `-p` option to webpack, which increases the minification of the bundle.


kshefchek:
+1

yuanzhou:
@DoctorBud thanks for the fix. I did a quick test and could see the fonts. I saw these font warnings in chrome console:

capture

PR #1136, opened 2015-12-22
Add functional tests for wormbase and flybase phenotype IDs in owlsim

kshefchek:
Adds a functional tests for getting phenotypes from golr given a pattern, and testing them in owlsim using the monarch API.

This is a brute force approach and takes a while (2-3 hours).  These only need to be ran once per new source.


DoctorBud:
Looks fine to me. +1

PR #1139, opened 2015-12-29
Fixes for cytoscape labs pages

kshefchek:
Revives cytoscape labs pages to do some preliminary testing on an "evidence" dag view that would work with the golr tables.  As of the initial PR, the /labs/cy-explore-demo page is working.  The /labs/cy-path-demo page does not work as we no longer have /graph/paths/\* queries available in scigraph.  I may use this route to test generation of evidence graphs.


kshefchek:
This could be merged in if there’s any interest in reviving the labs/cy-explore-demo page.  There are likely better options at this point in terms of cytoscape layouts (or d3), but this is a starting point.

cmungall:
+1

PR #1143, opened 2016-01-11
Ready to Merge - Adds Checkbox-style Facet Filters. Other UI cleanup. Removes concatenated monarch.js

DoctorBud:
- Adds Facet Filters using checkboxes, and an All and None button. Uses facet_filters.js, a variant of live_filters.js.
- The file monarch.js and its ilk (.min.js and so on) are now obsolete.
  The concatenation of 5 source files has been replaced by a CommonJS
  module format file for each.
- ESLint compatibility has been improved, but is still disabled in many source files.
- Fixes install.sh and readme to not use -p with webpack
- Cleans up some visual artifacts from the Tabs within the gene/disease/etc specific pages.
- Changes the Navbar to be a little more compact and changes the collapse threshold to a smaller value so that the navbar stays uncollapsed longer.
- Deletes golr-table-filter.mustache and makes golr-table.mustache behave with an optional filter.


kshefchek:
+1, the new filters are great!

PR #1144, opened 2016-01-11
Cleanup source, reduce asset bundle size, make Travis tests pass via stage

DoctorBud:
- Tunes the webpack configuration to massively reduce the bundle size and thereby the browser load time for initial loads.
-  Current bundle size is under 1Mb, with another 1Mb of fonts, images, sourcemaps and such.
- Enables cache-busting via webpack naming the bundle and chunk files with a hash.
- Various cleanup of dead code. Remove debug messages.
- Update monarch.js legacy bundle via 'make bundle'
- Adds a scripts/runWebpackDevServer.sh file which is invoked via 'npm run dev' and manages some of the issues with cache-busting filenames. 
- In addition to 'npm run dev', you can do 'npm run dev production', which will use the production environment.
- Similarly, you can do 'npm run bstart production' to exercise the bundled app against the production environment (since beta is down right now, this is helpful).
- Modifies .travis.yml to use 'stage' environment.
- Disables the phenogrid index.html example from the maui_phenogrid.feature test, because it is hardcoded to use beta. Increases minification of bundle by enabling variable name mangling (this may introduce bugs, all tests pass).


kshefchek:
+1

PR #1157, opened 2016-01-27
updates js-yaml to latest version in package. fixes conf/monarch-teamyaml to be correct YAML (there were some ambiguities and errors). Updates README to reflect that Bundled Mode is now the default

DoctorBud:
- updates js-yaml to latest version in package.
- fixes conf/monarch-team.yaml to be correct YAML (there were some ambiguities and errors).
- Updates README to reflect that Bundled Mode is now the default


frdougal:
+1, I approve this message

PR #1180, opened 2016-02-18
Replace autogenerated golr json schema with gulp tasks

kshefchek:
The current process to generate a golr conf file is:
1. Merge tab yaml "stub" files with a reference (oban-config) with MergeYamlConf.pl
2. Convert and merge all yaml files into a single json file with confyaml2json.pl
3. Commit these generated files to our repo (sorry)

I've replaced this with almost equivalent gulp tasks, the diff between the old and new files does contain some small differences:
1. "true" and "false" are now true and false
2. integers are no longer quoted

Also, wave goodbye to disco turtle

turtle


jmcmurry:
Boom. +1 @kshefchek

jmcmurry:
Also awesome, this: 
screen shot 2016-02-18 at 1 02 07 pm

PR #1183, opened 2016-02-21
Improve charite

iamkoehler:
Hi, not sure this is correct. I created a new branch and updated 3 minor things for the team page:
- change charite-link
- enable the picture next to my name
- update my picture

Let me know if this is the correct way to suggest such changes?


kshefchek:
+1

PR #1187, opened 2016-02-25
DO NOT MERGE. Fixes issue 1160 such that a POST to a redirected (30x) URL work.

DoctorBud:
- Fixes issue 1160 such that a POST to a redirected (30x) URL will magically be redirected. By default, the NodeJS request() package will redirect GETs.
- Also adjusts the server_config_production.json and server_config_beta.json to use the new OWLSim URLs (owlsim.monarchinitiative.org) which then redirect to monarchinitiative.org/owlsim

### DIFF
--- a/conf/server_config_beta.json
+++ b/conf/server_config_beta.json
@@ -9,7 +9,7 @@
 
     "golr_url" : "http://solr.monarchinitiative.org/solr/golr/",
 
-    "owlsim_services_url" : "http://monarchinitiative.org/owlsim",
+    "owlsim_services_url" : "http://owlsim.monarchinitiative.org",
 
     "analytics_id" : ""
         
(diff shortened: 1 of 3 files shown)


cmungall:
didn't check travis, but a +1 for the PR

DoctorBud:
Turns out that my fix works for GET, but that the `request` library we are using for making calls has a bug in it with respect to redirected POSTs. I've reported it here (it is easily fixed):
- https://github.com/request/request/issues/2060

Workarounds:
- I can convert the POST to a GET, but then large POSTs will get truncated by the conversion and will fail.
- I can explore alternative libraries to `request`, which may handle this POST redirect better. Probably easy to build a Node script to exercise this.
- I can fix the request bug, build a custom library, make it available somehow (npm distribution), and we'd just use that.

I'm going to wait until next week to see if the author of `request` responds to my comments. This PR is on hold for now and I'm leaving the config files alone.

DoctorBud:
Closing this. We no longer do a redirect on POSTs, so this is moot for now.

Issue #1019, opened 2015-10-16
literature/pub tabs


nlwashington:
add literature/pub/citation tabs to each of the main pages:
- [ ] disease
- [ ] gene
- [ ] genotype
- [ ] phenotype
- [ ] variant

these can be developed based on the literature-\* golr.

kshefchek:
Currently blocked by not having the golr index( @jnguyenx working through disk space issues ).

nlwashington:
How's this coming along?  Would be very keen to show this off to some folks.
@kshefchek are you able to make a workaround to add info about the pubs?  author/title/year/etc?

nlwashington:
i'd prioritize phenotype/disease pages, if possible.

cmungall:
It seems this is virtually done

We have pages
https://monarchinitiative.org/literature/PMID:7852544

The assoc tab doesn't work but AFAICT everything should be in golr. E.g. https://github.com/monarch-initiative/configs/blob/master/SciGraph/golr/queries/literature-disease.yaml

Maybe best to wait til after @DoctorBud does any refactoring to consolidate mustache templates?

For comparison:
http://amigo.geneontology.org/amigo/reference/PMID:22176531

I guess we ended up writing similar eutils code. My preference would have been for a common lib shared in npm, at least for the api-client part (surely biojs have done this), cc @kltm, but not a big deal.

kshefchek:
I can fix the association tab and start linking pubs internally, although I've only tested the page on pubmed IDs.  For publication tabs, we need to implement the indexing of title, author, and pub info in solr.  We can either do this as a post processing step after creating our RDF files, or this could be part of the dipper ingest, as a module that runs at the end of each ingest.

See https://github.com/monarch-initiative/monarch-app/pull/1192 for more info.

jmcmurry:
@kshefchek note that we are currently hyperlinking literature references from the spotlights (see from /phenotypes for example), but the destination is blank within the page. We should either turn off that hyperlink from spotlight or make them work. Latter obviously better, but harder.

Issue #124, opened 2014-02-11
Literature pages


nlwashington:
We need 'literature' pages.

These should have tabs just like our other pages summarizing the mentioning of:

*models (asserted)
*genotypes
*diseases
*phenotypes
*pathways
*genes
*related publications -- this last one will be interesting... we can do the easy thing of "other publications that mention X (like genes, or same author, or whatever), but a novel idea is related publications based on phenotype!

*analyze -- our analyze functions ought to let us "search" for publications based on phenotypes too.  how cool is that?

nlwashington:
If we want better search, say by author, we will need NIF to include authors in their literature indexing.  See JIRA ticket: https://support.crbs.ucsd.edu/browse/NIF-9854

harryhoch:
see also https://github.com/monarch-initiative/monarch-app/issues/175

nlwashington:
Literature indexing now includes authors and grant ids:

http://beta.neuinfo.org/services/v1/literature/pmid?pmid=22080565

nlwashington:
These would have the form 
http://www.monarchinitiative.org/literature/PMID:22080565
There are a couple of ways we could "label" that identifier.  My initial thought is to just use the title of the article.  And the jumbotron should include a short citation (first author et al (year) issue:pages) as a subtitle.
The basic information about an article in the first tab should be populated with data acquired by the literature service, 
http://beta.neuinfo.org/services/resource_LiteratureService.html as described in #123.  
Which should show authors, title, short journal title (issue, pages, etc), mesh terms, and abstract.

nlwashington:
We can initially populate a "similar papers" table that uses the /literature/moreLikePmid service.  This would be very cool.

nlwashington:
@ccondit do you do any indexing with the ontology for literature services?  for example, it appears i can do this query:  http://beta.neuinfo.org/services/v1/literature/search?q=HP:0000478, but i don't think it's actually matching on 'abnormality of the eye', which is the ontology term that it's matching to.  have you considered this?

ccondit:
Negative - that query is not correctly interpreted (it's just finding the numeric portion, I believe). Are you trying to perform closure queries on the literature index?

nlwashington:
yes, that was my thought.  is that possible and/or good idea?  it might seem that it would have to go through some kind of annotation layer first, right (which could certainly provide many false positives)?

nlwashington:
[...]

PR #1192, opened 2016-02-26
Update literature pages and add lit tabs

kshefchek:
This PR is focused on updating literature pages and adding literature association tabs to other pages (disease, phenotype, etc.)

Currently we have legacy literature pages driven from NIF views, but these are not linked internally from anywhere on the site, for example: http://monarchinitiative.org/literature/PMID:15459382

The goals of this PR are to:
- [x] Make literature pages driven from eutils and golr
- [ ] Add literature tables to other pages driven by golr and eutils
- [ ] Link publications in all tables to new literature page

This is a slight departure from our current golr widget in which UI features (paging, filters, etc) are coupled to a single golr call.  Here we would pull publication-entity associations from golr and then get additional information from eutils.  There is also some post processing required to uniquify and remove poorly formed pub ids.

This could push forward development of solr cross poduct views, such as orthology associations + orthology phenotype counts (or profile similarity from owlsim).

Related tickets:
#124
#1019

https://github.com/monarch-initiative/dipper/issues/267

Watchers: @nlwashington @jmcmurry


kshefchek:
I've fixed up the literature pages to be driven from eutils and golr.  However, the individual tabs don't quite make sense here and have many duplicates.  I think an improvement would be a single associations table with subject-object associations where a publication is either direct or indirect evidence - and having this qualifier value on the table as well.

kshefchek:
We've decided to index publication information in solr, rather than pull this information from eutils in the client and perform a solr and eutils join.  As this PR stands, it adds a monarch literature page, but it is not linked anywhere internally.

Note there was also some work done to improve the filter widget, but I have never had the bandwidth to finish.  Instead of leaving this PR open indefinitely, I've decided to roll back the changes to the client widgets and break this out into a new branch. The majority of changes from bf1ba51 on have been rolled back.

DoctorBud:
+1 Looks good. Tests passed on my machine except for the usual suspects.

Issue #1196, opened 2016-03-01
Search result failure for uticaria


jmcmurry:
A friend's kids have [cold uticaria](http://monarchinitiative.org/disease/OMIM%3A120100) so I was curious to know if I could find it in Monarch. There was no autocomplete suggestion for uticaria so when I just pressed enter to search I got this, which I think we can all agree is bad, but I'm having a hard time working out what could be behind it. 

http://monarchinitiative.org/search/uticaria

<img width="622" alt="screen shot 2016-02-29 at 9 22 57 pm" src="https://cloud.githubusercontent.com/assets/3658863/13418507/d40f489c-df2a-11e5-86c5-fd6038c789d4.png">

If I search for "cold" and I wait very very patiently, I do get "cold uticaria familial" showing up in autocomplete. I also get sensible autocomplete results.

Thoughts?

kshefchek:
SciGraph autocomplete/search isn't great with misspellings.  We have a suggestions service that we could use here as a patch:
http://scigraph-ontology.monarchinitiative.org/scigraph/vocabulary/suggestions/uticaria?limit=5

Oddly enough the correct spelling doesn't return the right suggestion:
http://scigraph-ontology.monarchinitiative.org/scigraph/vocabulary/suggestions/urticaria?limit=10

jmcmurry:
I guess we should always return some kind of result, regardless, so if we handle the misspelling, the issue of handling "no results" will go away.

kshefchek:
I'll add suggestions in from SciGraph, but we should also add as a TODO for the new solr based autocomplete and search, cc @cborromeo and @harryhoch.

PR #1197, opened 2016-03-01
Add suggestions when search result not found

kshefchek:
Adds suggestions from SciGraph when a search result is not found.  This is a small enhancement while a more robust search/autocomplete widget is built with solr.

See #1196 

suggest


lib/monarch/web/webapp.js
```diff
@@ -1076,10 +1076,24 @@ function searchHandler(request, term, fmt) {
...
             info.resultsTable = function() {return genTableOfSearchResults(info.results); };
             info.description = "<span>Results for "+term+ " searching phenotypes, diseases, genes, and models</span>";
         } else {
             info.resultsTable = ''; // "<span class=\"no-results\">&nbsp;&nbsp;No results found</span>";
+            info.hasResults = false;
+            var suggestions = engine.getSciGraphSuggestions(term, 5);
+            var suggestHead = "<h4>No results found, did you mean:</h4>";
+            var suggestSpan = '';
+            suggestions.forEach( function (i) {
+                var href = "<a href=\"/search/" + i + "\">" + i + "</a>";
```
DoctorBud: style/readability: When writing suggestions.forEach, I think it is more understandable if the loop variable is named the singular of the collection variable. So in this case `i` should really be `suggestion`.

DoctorBud: Do we know that `i` will always be in a URL format? Should it be URIEncoded prior to using it to build the `/search/...` url?

DoctorBud:
Looks OK, provided that the URIEncoding issue above is considered. Provisional +1

lib/monarch/web/webapp.js
kshefchek: Good point, using function (i) is a bad habit, especially since it's not an index variable.

PR #1205, opened 2016-03-15
Provide global solr variable to client

kshefchek:
Since we are moving to a solr backed autocomplete/search widget, we need to provide the solr url and golr conf files to all pages.


frdougal:
+1

kltm:
Keep in mind that a lot of this will/can be changed as I reach completion of https://github.com/geneontology/amigo/issues/317 and https://github.com/berkeleybop/bbop-widget-set .

PR #1211, opened 2016-03-24
Adds HPO Cameo area on home page. Other minor cleanups.

DoctorBud:
- Adds HPO Cameo area on home page.
- Removes duplicate eslint.
- Removes unnecessary variables.less comments in monarch.less.
- Disables 2 of 3 maui_phenogrid.feature tests due to issues with the Phenogrid documentation page (internal Phenogrid works fine, it's only the Documentation example that is busted).
  Note: This PR relies upon the new toolchain and the use of the LESS preprocessor to customize Bootstrap styles. Running in old, non-webpack, mode will work, but will look wrong.


kshefchek:
+1

Issue #1200, opened 2016-03-07
Add phenopacket exporter


cmungall:
Each page/tab should have an 'export phenopacket' button (similar to the old export json)

TBD: exporter at the level of golr or scigraph

DoctorBud:
If we allow for the selection of a subset of rows (e.g., phenotype rows within a disease page), then we'll need to enable per-row checkboxes and have the Export button apply to selected rows. This feature exists as an option within the BBOP widget we use, but has been unused for a while and Facet Filters may have ignored support for them.

kltm:
As long as the manager is used (properly), there should be no issue.

kshefchek:
I am thinking variant-phenotype associations would be a good candidate as an example; however, when decomposing these associations they typically fall under the pattern of:
1. Variant inferred to have a phenotype through a disease-phenotype relation
2. Variant inferred through a genotype-phenotype relation

Is it possible to represent this in the level-1 schema, through a  series of 2 or more associations, or am I misunderstanding this portion of the schema?  We could alternatively export genotype-phenotype direct relationships, and variant-disease direct relationships.

cmungall:
I suggest an iterative approach: let's focus on simple variant-phenotype 
associations for a first pass, with minimal evidence

Then let's do another pass, and we can explore extension of the evidence 
model, bringing in the work that you and @mbrush have been doing

PR #1212, opened 2016-03-25
Add phenopacket export button/endpoint to monarch-app

kshefchek:
See #1200 

This is an experimental, untested branch for exporting data from golr/scigraph into the phenopacket format.  As a first pass just implementing for variant-phenotype tables/golr views.  The approach I have taken is to get the current state of the golr manager (filters, query) and send this back to a monarch rest service that constructs the phenopacket.  For example:

```
    http://localhost:3000/phenopacket?q=*:*&fq=subject_closure:%22ClinVarVariant:186784%22&fq=object_category:%22phenotype%22&personality=variant_phenotype
```

I've used some features of ES6 and would require us to update node to the latest recommended version 4.4.1.  I've tested this both locally and on beta and things seem to be running smoothly, although I have not fully vetted that we have maintained compatibility with all our dependencies.

https://beta.monarchinitiative.org/status

To test locally:
nvm install 4.4.1
nvm use 4.4.1

and then change the solr url to solr.monarchinitiative... in the dev config file.


cmungall:
+1

kshefchek:
I've added a beta badge to the button, as well as all variant types from the closure map with SO or GENO prefixes.

PR #1215, opened 2016-03-30
change app base to beta.monarch on beta.monarch

kshefchek:
Changes app_base configuration on beta.monarchinitiative to beta.monarchinitiative, this was possibly changed when we were making the switch to https or for phenogrid.  @yuanzhou @jnguyenx can this be switched back?

### DIFF
--- a/conf/server_config_beta.json
+++ b/conf/server_config_beta.json
@@ -1,7 +1,7 @@
 {
     "type" : "beta",
     
-    "app_base" : "https://monarchinitiative.org",
+    "app_base" : "https://beta.monarchinitiative.org",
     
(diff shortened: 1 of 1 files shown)


yuanzhou:
@kshefchek just to be clear, you want me to point to `https://beta.monarchinitiative.org` in Phenogrid?

kshefchek:
This is just for the monarch-app, I thought it might be related phenogrid but forgot this is set in a different configuration.

yuanzhou:
@kshefchek thanks for clarifying. Is it OK if I only use https://monarchinitiative,org everywhere?

PR #1248, opened 2016-05-16
Fixes bug where blog data would not be updated properly

DoctorBud:
Fixes bug where blog data would not be rechecked every 30 minutes, because each user load of our landing page would reset the timer. The fix is to only reset the timer when we check blogspot.

### DIFF
--- a/lib/monarch/web/webapp.js
+++ b/lib/monarch/web/webapp.js
@@ -3399,15 +3399,18 @@ var _blog_data_last_res = [];
  * Get blog data
  * @param {String} label - An optional label restricting blog posts.
  */
+
 function _get_blog_data(label) {
 
   // Only proceed if the time-since-last threshold is crossed or the
   // cached results are empty.
   var res = [];
-  var now = _get_now_sec()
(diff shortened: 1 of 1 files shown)


DoctorBud:
Here’s what was happening when a user hits the landing page:
- We re-check the blogspot site if 30 minutes has elapsed since we last updated the _blog_data_last_time variable
- If less than 30minutes has elapsed, we simply return the ‘old’ results and avoid hitting blogspot
- If more than 30 minutes has elapsed, we fetch new data from blogspot
- The error is that we unconditionally update _blog_data_last_time after each user request, rather than updating it after each blogspot request.

The result will be that if our landing page is visited more frequently than once per 30 minutes, we will NEVER recheck blogspot.

The error was probably introduced months ago during editing (might have been my error), where the setting of _blog_data_last_time was moved outside of a conditional’s curly braces.

I simulated the bug by turning down the refresh time to 20 seconds, where I was able to inhibit the checking of blogspot simply by refreshing more often than 20 seconds.

PR #1255, opened 2016-05-21
UI Optimizations. Logging cleanup. Font-delivery fixes. Twitter workarounds.

DoctorBud:
- Updates MGI source logo to use the latest from http://www.informatics.jax.org/webshare/images/mgi_logo.gif
- Improves look of Twitter timeline widget by removing the new header/footer and the new image preview.
- Tunes .eslintrc to make some old Javascript constructs warnings
  instead of errors.
- Cleans up webapp.js to be more ESLint-friendly.
- Deletes unnecessary web initialization console messages.
- Fixes webenv.js to encode SFNT, WOFF, and WOFF2 using correct mimetype
  and binary encoding.
- Reduces size of hpo-apert-cameo.png to save bandwidth.
- Deletes 'npm run brun' since it is unnecessary.
- Removes stack trace debug message when a page not found is generated.
- Removes other console log messages.


kshefchek:
+1

Issue #1259, opened 2016-06-01
Mouse genes don't have taxon in autocomplete and search


kshefchek:
This is a bug from autocomplete and search running from the ontology scigraph, and the taxon completion calling the data scigraph.  It only affects mouse genes because MGI overrides NCBIGene as the cliqueleader in the data scigraph, and therefore taxon information is hanging off the MGI nodes rather than NCBIGene, which are always clique leaders in the ontology scigraph because we only load ncbigene.ttl

Should be a quick fix.

PR #1260, opened 2016-06-01
fix autocomplete mouse issue

kshefchek:
Fixes autcomplete and search issue where mouse taxon does not show up, see #1259 
Also
- increased search limit to 500, autocomplete to 30
- prioritizes common models

These are just patches to hold us over until the solr backed search/autocomplete overhaul.

### DIFF
--- a/js/monarch-common.js
+++ b/js/monarch-common.js
@@ -318,10 +318,34 @@ function add_species_to_autocomplete(data, map, gene_ids) {
             if (map[i]['id'] == id) {
                 if (label) {
                     map[i]['tag'] = label;
+                    if (label == 'Man') {
+                        map[i]['tag'] = 'Human'
+                    }
                 }
             }
         }
     });
+    // Sort by the common taxa
+    var temp_map = [];
+    temp_map = temp_map.concat(map.filter( function (i){
+        return i.tag == 'Mouse' })
+    );
+    temp_map = temp_map.concat(map.filter( function (i){
+        return i.tag == 'Zebra Fish' })
+    );
+    temp_map = temp_map.concat(map.filter( function (i){
+        return i.tag == 'Human' })
+    );
+    temp_map = temp_map.concat(map.filter( function (i){
+        return i.tag == 'Fruit Fly' })
+    );
+    map = temp_map.concat(map.filter( function (i){
+        return ((i.tag != 'Mouse') 
+              && (i.tag != 'Human')
+              && (i.tag != 'Fruit Fly')
+              && (i.tag != 'Zebra Fish'))
+        })
+    );
     return map;
 }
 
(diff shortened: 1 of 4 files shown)


DoctorBud:
@kshefchek I played around with this yesterday (didn't know you were working on it). I upped the autocomplete limit to 100 and made it so that the menu scrolls. If you want to try it out, modify `monarch-common.css` from:

```
#monarch-body .ui-autocomplete {
    position:absolute;
    line-height: 22px;
    cursor:default;
    white-space: nowrap;
    font-size: 100%;
    color: #888888;
    z-index:8000 !important;
    padding: 0;
    padding-right:15px;
    background: #DDD;
    list-style-type: none;
    text-decoration: none !important;
}
```

to (adding 3 lines):

```
...
#monarch-body .ui-autocomplete {
   ...
    text-decoration: none !important;
    /* Scrollable autocomplete menu adds below */
    overflow-x: hidden;
    overflow-y: auto;
    bottom: 20px;
}
```

The only downside is that when there are not enough items, the menu is bigger than necessary. Feel free to adopt my changes in addition to your more semantically-oriented changes, or not. Just wanted to point out that a scrolling dropdown is possible and sometimes useful.

Also, there are a few `console.log` that should be removed from `search_form.js`, since they bloat the browser log and slow things down a bit:

```
--- a/js/search_form.js
+++ b/js/search_form.js
@@ -54,7 +54,7 @@ function navbar_search_init(in_search_id, in_form_id){
                        collision: "flip"
            },
            source: function(request, response) {
-               console.log("trying autocomplete on "+request.term);
+               // console.log("trying autocomplete on "+request.term);

                // Argument response from source gets map as argument.
                var _parse_data_item = function(item){
@@ -83,7 +83,7 @@ function navbar_search_init(in_search_id, in_form_id){
                    };
                };
                var _on_success = function(data) {
-                       console.log('success:', data);
+                       // console.log('success:', data);
                    // Pare out duplicates. Assume existence of 'id'
                    // field. Would really be nice to have bbop.core in
                    // here...
@@ -137,7 +137,7 @@ function navbar_search_init(in_search_id, in_form_id){
                };

                var query = "/autocomplete/"+request.term+".json";
-               console.log('about to do query:', query);
+               // console.log('about to do query:', query);
                jQuery.ajax({
                    url: query,
                    dataType:"json",
```

DoctorBud:
If you adopt my CSS changes above, you will also need to change the padding on that style to:

```
    padding-right:15px;
```

So that when the vertical scrollbar appears, there will be enough padding to avoid the dreaded horizontal scrollbar.

DoctorBud:
+1 to your changes, in any case.

kshefchek:
I'll hold off on the scrolling autocomplete for this PR but like the concept, removed comments in search_form.

Issue #1277, opened 2016-06-10
Clean up hover text for column headers


kshefchek:
Many of our column headers contain incomplete textual descriptions.

For example:
Evidence Type: has a question mark at the end
Three fields contain TBD notes - source, qualifier, and and assigned by, although only source is shown in our tables.

Note this adjusted by editing the YAML files here:
https://github.com/monarch-initiative/monarch-app/tree/master/conf/golr-views

jmcmurry:
@kshefchek is this closable now?

PR #1278, opened 2016-06-12
add worm to phenogrid config, update some header text see #1277

kshefchek:
Adds C. elegans to phenogrid config, update some header text see #1277


kshefchek:
This PR only includes changes to config and test files, the only code change was reverted.  Merging it in.

PR #1286, opened 2016-06-23
Analyze phenotypes - from zhou

yuanzhou:
This pull request contains many small UI changes based on the discussion at https://github.com/monarch-initiative/monarch-app/issues/1250. As I explained to @jmcmurry, it's considered as a quick and dirty fix for the release purpose.

There are still couple things need to be addressed after the release:
- Fix gene autocomplete bug
- Add species to filter out phenotypes when creating the phenotype profile
- Make Step 2 predefined target groups checkboxes instead of dropdown menu

@harryhoch @kshefchek please review and merge the PR if it looks good. Thanks!

PR #1288, opened 2016-07-07
Fixes/Updated for GSA release

kshefchek:
- Remove direct/indirect from relationship column
- Fix mgi_slim and zfin_slim logos

TODO
Add ortholog-phenotype tab

### DIFF
--- a/lib/monarch/api.js
+++ b/lib/monarch/api.js
@@ -7959,6 +7959,38 @@ bbop.monarch.Engine.prototype.fetchPhenotypes = function(id) {
     return phenotype_list;
 };
 
+//Hack to get patient pages, will eventually go away in favor of getting this info from a golr document
+bbop.monarch.Engine.prototype.fetchCliqueLeader = function(id, server) {
+    var engine = this;
+    var leader = null;
+    var path = "dynamic/cliqueLeader/"+id+'.json';
+    var resultObj = [];
+    
+    if (!server){
+        server = engine.config.scigraph_data_url;
+    }
+
+    var uri = server+path;
+    var exchangeObj = engine.fetchUrlWithExchangeObject(uri);
+    
+    if (exchangeObj.status == 404) {
+        this.log("not found: "+uri);
+        return {concepts:[],nodes:[],edges:[]};
+    }
+    if (exchangeObj.status == 404) {
+        // TODO: throw error
+        return errorResponse(exchangeObj.content);
+    }
+    var sciObj = JSON.parse(exchangeObj.content);
+    var graph = new bbop.model.graph();
+    graph.load_json(sciObj);
+    var node_list = graph.all_nodes();
+    if (node_list.length == 1) {
+        leader = node_list[0].id()
+    }
+    return leader;
+}
+
 // Basically a copy paste of fetchClassInfo that hits the SciGraph data server isntead of the ontology
 // This should all be refactored into one function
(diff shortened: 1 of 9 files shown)


lib/monarch/api.js
```diff
@@ -7959,6 +7959,38 @@ bbop.monarch.Engine.prototype.fetchPhenotypes = function(id) {
...
+bbop.monarch.Engine.prototype.fetchCliqueLeader = function(id, server) {
+    var engine = this;
+    var leader = null;
+    var path = "dynamic/cliqueLeader/"+id+'.json';
+    var resultObj = [];
+    
+    if (!server){
+        server = engine.config.scigraph_data_url;
+    }
+
```
DoctorBud, line 7972: Just an FYI, from what I've been seeing in contemporary JS code, I've adopted the following more concise style for this type of `if` statement:

```
server = server || engine.config.scigraph_data_url;
```

Where the short-circuited disjunction will only evaluate (and return) the second expression if the first is not 'truthy'.

The trick only works in weakly-typed languages (like Python and JS), but it appears to be a common way to write it.

lib/monarch/web/webapp.js
```diff
@@ -4286,6 +4296,13 @@ function variantByIdHandler(request, id, fmt) {
             var newID = id.replace("_",":");
             return web.wrapRedirect(genURL('variant',newID));
         }
+        
+        // OMIM variants have a period in the ID, add back to id and
+        // set format to null, TODO account for actual .json format
+        if (id.match(/^OMIM/)){
```
DoctorBud, line 4302: As of ES6, we can use `str.startsWith` for a more readable expression. I think our Babel ES6 compiler deals with it in the browser with a polyfill, and that Node 4+ supports it too. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith

Until writing this post, I assumed startsWith/endsWith were standard, but apparently only as of ES6

DoctorBud:
Looked it over. Looks fine to me.

+1 (I have no comment on the failing tests)

kshefchek:
The same test fails on master, @yuanzhou could a phenogrid update have broken this:
And I wait for id "phen_vis_svg_group"

kshefchek:
Looks like it is failing because BLOC1S3 is no longer associated with renal insufficiency, looking into it now.

kshefchek:
Looks like this is not a bug, but from fixing our gene-phenotype query to not propagate phenotypes from gene-disease relationships where a gene is a marker/mechanism for a disease. Will fix the behave test.

yuanzhou:
@kshefchek when there's no "BLOC1S3" associated with renal insufficiency, Phenogrid might have output a message instead of the grid rendering. I also tried to match that error message or even tried different div ids but all failed.

One thing I noticed is when there's no grid rendering, there's no id "phen_vis_svg_group" so I totally removed that "wait for id" line (https://github.com/monarch-initiative/monarch-app/pull/1286/commits/b0afa921b5806b8a410b5a14d24b8c8fed40fdf3). Still failed. Even after I commented out that test (https://github.com/monarch-initiative/monarch-app/pull/1286/commits/11dca029b351b7f01a9c12f094b596c880488bdb). If you see this build history, https://travis-ci.org/monarch-initiative/monarch-app/builds/143150102, it's a different error.

screen shot 2016-07-07 at 8 56 58 pm

PR #1307, opened 2016-07-12
Fixes Feedback window, which has been broken since January.

DoctorBud:
##### Fixes Feedback window, which has been broken since January.
- Adds a scripts/relay-feedback.sh script which can be used to copy the feedback file to a central location.
- Fixes styling of close boxes on feedback dialog and response alert.
- Adjusts position of alert to be visible.
- Removes 'ListServ' reference from feedback window in favor of a link to a Monarch Survey and signup mechanism at https://bit.ly/monarchcommunity

### DIFF
--- a/css/monarch-common.css
+++ b/css/monarch-common.css
@@ -984,8 +984,8 @@ a:hover .hilite, a:active .hilite {
 }
 
 .feedback-close {
-  margin-top:-2px;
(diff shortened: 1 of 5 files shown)


kshefchek:
+1

jmcmurry:
Thanks; looks great :)

PR #1315, opened 2016-07-25
fix taxon bug in compare call

kshefchek:
Fixes issue with gene taxon not showing up in compare results

### DIFF
--- a/lib/monarch/api.js
+++ b/lib/monarch/api.js
@@ -1454,18 +1454,19 @@ bbop.monarch.Engine.prototype.filterXrefsForGene = function(geneMap,references) 
 bbop.monarch.Engine.prototype.fetchTaxon = function(id) {
     var engine = this;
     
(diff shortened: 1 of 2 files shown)


jmcmurry:
yay! this is huge, thanks

frdougal:
@kshefchek  That's great.  Currently I can retrieve the taxon for Homo sapiens genes only.  The taxon is stored as an object with two attributes: id (NCBITaxon:9606) and label (Homo sapiens).  Will your changes follow the same approach?

kshefchek:
@cborromeo correct, let me know if it doesn't work.

frdougal:
@kshefchek can I use the non-production servers for solr?  I think Jeremy was still working on the servers.  Can I access them using the connections in server_config_dev.json?

kshefchek:
@cborromeo it's probably easier to continue to develop using the production servers for now.  Hopefully we can unpassword protect the dev servers soon.

Issue #975, opened 2015-09-30
Clear out the cruft in api.js


cmungall:
There are vast swathes of api.js that are unused. Anything that calls anything that uses federation_services_url can go

caveat: landing pages still have a dependency. @kshefchek had the idea of just checking in the json cache for these as a temporary solution.

jmcmurry:
@kshefchek any update? What exactly would be needed to close this?

kshefchek:
I've taken the approach of commenting out portions and then deleting if all seems fine.  In the literature pages PR I commented a good chunk of the api.js code, and at this point it's probably safe to delete it.  That said there's still a bit more than can be cleaned up on top of this.

cmungall:
Yes, delete commented out code. This is good practice anyway, we have memory in git.

I'm not sure how easy it is to see a call tree in javascript, but `fetchDataFromResource` must die, and along with it any code that calls it. 

Similarly, any method name with ontoquest in it.

Note the landing pages no longer have a dependency, so this should be much easier

We can also kill anything for ringo

Although this will not affect functionality, it's important for understanding our own codebase and moving forward with future refactors.

I don't think it should take long

cmungall:
I'll do this

cmungall:
@kshefchek is generateHistogram used? It has a bunch of dependencies to federation

cmungall:
I believe these are no longer used:

```
web.wrapRouteGet(app, '/gene/:id/:section.:fmt?', '/gene/{id}/{section}.{fmt?}', ['id', 'section', 'fmt'], fetchFeatureSection);
web.wrapRouteGet(app, '/variant/:id/:section.:fmt?', '/variant/{id}/{section}.{fmt?}', ['id', 'section', 'fmt'], fetchFeatureSection);
web.wrapRouteGet(app, '/genotype/:id/:section.:fmt?', '/genotype/{id}/{section}.{fmt?}', ['id', 'section', 'fmt'], fetchFeatureSection);
web.wrapRouteGet(app, '/model/:id/:section.:fmt?', '/model/{id}/{section}.{fmt?}', ['id', 'section', 'fmt'], fetchFeatureSection);
```

I will get these on the next pass

kshefchek:
@cmungall no it has been replaced, feel free to remove.

Issue #1291, opened 2016-07-11
Selenium, xvfb and Firefox `behave` tests need to be aware of window size


DoctorBud:
When we execute the `behave` unit tests in `monarch-app/tests/behave/` in the TravisCI environment, the browser emulation is carried out with Firefox, `xvfb` and Selenium. One of these components appears to be very strict about which DOM elements are visible to the web driver. For example, an element `phen_vis_svg_group` that is outside of the virtual window's bounds will not be detected when the `behave` tests use the clause `and I wait for id "phen_vis_svg_group"`

This behavior came up previously with the Analyze page, but recent UI layout changes have caused it to resurface. The older issue is https://github.com/monarch-initiative/monarch-app/issues/1017

I recently made the following PR to allow the `maui_analyze.feature` test to complete when run via TravisCI: https://github.com/monarch-initiative/monarch-app/pull/1289

I have been unable to find documentation that describes this limitation/feature of `xvfb`, but there is evidence that other users have experienced this type of problem. I've included some of the most relevant ones below:
- http://sqa.stackexchange.com/questions/15459/when-running-selenium-tests-on-another-machine-page-element-renders-offscreen-s
- https://github.com/seleniumhq/selenium-google-code-issue-archive/issues/5828

PR #1327, opened 2016-08-10
Remove cruft

cmungall:
Clearing cruft that I am 99% sure is unused, see #975


DoctorBud:
+1
It's satisfying to see dead code go away.

Not sure what's going on with the failing tests, but both of the failures are in code that is loading a Phenogrid. There are two modes of failure that I've observed:
- The data URL that Phenogrid is using is broken, so that Phenogrid cannot load. This happens when Phenogrid is pointing at the beta GOLR or SciGraph servers and they are down for some reason.
- The virtual window size is too short for the Phenogrid SVG element to be visible, so that Selenium/Firefox fails to find it. See https://github.com/monarch-initiative/monarch-app/issues/1291 for more info. In theory, you have my fix from this PR https://github.com/monarch-initiative/monarch-app/pull/1289

jmcmurry:
@harryhoch @yuanzhou could you please look into this? Thank you :)

harryhoch:
@yuanzhou, do you have a few minutes to review? Thanks

DoctorBud:
One thing that might be causing this is if Phenogrid is referring to one of the 'dev' SciGraph or GOLR servers that is currently under a username/password regime due to UDP data. Since Phenogrid accesses these servers without logging, we can't see that fact from the Travis logs. So it might be worth verifying that Phenogrid doesn't have any hardcoded server references in its code. All of PG's config should derived from the config that the monarch-app is using, currently:

```
Starting server. Platform: NodeJS. Environment: beta. Port: 8080.
NODEVERSION: v4.4.1
defaultConfigFile: conf/server_config_beta.json golrConfigFile: conf/golr-conf.json
defaultConfig: {"type":"beta","app_base":"https://beta.monarchinitiative.org","scigraph_url":"https://scigraph-ontology.monarchinitiative.org/scigraph/","scigraph_data_url":"https://scigraph-data.monarchinitiative.org/scigraph/","golr_url":"https://solr.monarchinitiative.org/solr/golr/","owlsim_services_url":"https://monarchinitiative.org/owlsim","analytics_id":""}
#USE_LOG_FETCH enabled
#USE_BUNDLE enabled
bbop.monarch.Engine Using pre-set configuration: 
bbop.monarch.Engine Using pre-set golr configuration: 
HapiJS Server running at: http://0.0.0.0:8080
```

If PG is pointing at https://scigraph-ontology-dev.monarchinitiative.org or https://solr-dev.monarchinitiative.org/, then it is incorrect (but this has happened in the past), and it will fail due to username/password.

yuanzhou:
Not sure if this is the cause, I went to https://beta.monarchinitiative.org/ and the login popup showed up to ask me username and password. It says 401 "Authorization Required".

kshefchek:
We had to change the monarch_base setting in the conf so phenogrid could access monarch app from the browser.  The build is fixed now.

PR #1352, opened 2016-09-06
Monarch Node Browser Simplification and Unification

DoctorBud:
I cleaned up many annoyances and inconsistencies in the Monarch UX. I
finally ended up with a single node.mustache template for all of the
basic nodes that Monarch displays. This will allow me to make changes in
a single location, providing a uniform version of the Overview tab and
the other node-associated lists.
- Unifies most of the leaf pages (disease, model, etc) into a single node.mustache template.
- Enables location-hash visibility and bookmarking.
- Simplifies the header line for node pages.
- Hides redundant columns corresponding to the current node and the association relationship (may make this hiding optional in the future).
- Increases default number of entries in a list to 25 from 10 and removes 10 from the set of choices.
- Deletes several legacy endpoints (need to delete their templates eventually).
- Updates several packages in package.json, but issues with jquery-ui and draggable and autocomplete made it so these versions couldn't be advanced until we fix our code.
- Fixes browsersync configuration so that changing a pup-tent-loaded JS will trigger a reload
- Adjust the MyGene launcher so that the description is located in the Description section of Overview.
- Ensure that 'npm run devp' disables puptent cache.
- Delete legacy code.
- Fixes Literature landing page so that it sort of works, rather than having a broken stylesheet.
- Sort XRefs so that the primary_xref is first.
- Move Species to the main title for Genes.
- Adjust behave tests to match UI changes.
- Reduce white space on landing pages and on specific (disease, model, gene, etc) pages.
- Move Annotation Score into the top pager area to reduce space waste.
- Clean up Annotation Score so that the instructions are displayed nicely and so that the Question Mark doesn't change position randomly.
- Shrink the size of the navigation tabs.
- Factor the AddGolrTable() calls into a common addNoneEmptyGolrTable() helper which ensures only non-empty tables are inserted, and which adds Filters if possible.
- Use skipFields if on a leaf node, to hide redundant space-wasting fields that are implied by the Page and Tab types (experimental feature that we can disable).
- Factor redundant subexpressions to tidy up code.
- Remove isparta package from package.json. It gives warnings and is unused.
- Fixes bug where dropdown menus on navbar are positioned too high.
- Improves narrow-width window responsiveness by ensuring that the Dropdown menus are displayed correctly. The Documentation menu previously was overflowing the dropdown container.
- Reduces height of collapsed navbar slightly.
- Reduces width of window where navbar collapse occurs from 875px to 840px
- Compresses layout of GOLR tables slightly, ensuring that Sources and References are fixed-width.
- Reduces font size of the Team Page title, and centers the title.
- Adds a bullet symbol to each evidence line, to increase readability
- Deletes unused and redundant version of process_entry from results_table_by_class_conf_bs3.js
- Modifies the Compare tab behavior so that Phenogrid is lazily loaded if the Compare tab is clicked or if the page is loaded to the Compare tab via the location hash. This results in faster page loads and was already a feature of the phenogridloader-onclick.js version of LoadPhenogrid().
- Deletes phenogridloader-no-species.js, because it was the same as phenogridloader-onclick.js, except that the no-species.js version was not lazily loaded.
- Ensures that the mygene description progress spinner is positioned within the target description area.
- Removes unnecessary makeSpinnerDiv() from browse.js. It turns out that we were adding a Spinner for each row, which were never displayed, but definitely bloated the DOM.
- Adds an Info icon to the Classification Browser that points to the monarch-disease-ontology and to the upheno ontology.
- Removes panel divs from around the dovegraphs, to reduce unnecessary borders and ink.
- Fixes irritating horizontal scroll on the landing pages due to the breadcrumbs region extending outside the window.
- Increases size of breadcrumbs by 10px to prevent text overwriting the breadcrumb and to increase readability.
- Updates dovegraph to 1.0.1.
- Updates README to indicate new minimum versions for NodeJS and npm.
- Ensure that the 'windmill' spinner on the landing pages is centered vertically within the loader pane.


cmungall:
So a generic template - will this be driven by the yaml schemas? do you need to coordinate with @jnguyenx ?

jnguyenx:
That's fine, I haven't touched the monarch-app code yet for autocomplete and search.

DoctorBud:
The template has always been mostly generic, with the only difference being a few lines in each of the node-specific templates (e.g., disease.mustache vs disease-group.mustache). I just factored out the deltas and use a mustache variable to contain the delta, rather than using a whole separate file.

This DOES put us on the road to having a generic `nodeByIdHandler`, as opposed to our `modelByIdHandler`, `geneByIdHandler`, etc. However, that is a subsequent refactor that would be fun and result in code deletion (yay!), but I haven't explored it in detail.

cmungall:
That makes sense @DoctorBud 

So a generic template would need to be driven by some kind of configuration, so that we know which columns to show and which fields to facet on. I assume here the configurations will be the yamls. This is essentially how amigo works.

js/lib/monarch/widget/display/results_table_by_class_conf_bs3.js
```diff
@@ -229,11 +229,13 @@ bbop.monarch.widget.display.results_table_by_class_conf_bs3 = function(cclass,
     }
     var results_order = cclass.field_order_by_weight('result');
     var headerColumns = [];
-    var skipFields = ['subject', 'relation']; // , 'subject_taxon'
-    if (handler.golr_field === 'object_closure') {
-      skipFields = ['object', 'relation']; // , 'object_taxon'
+    var skipFields = [];
```
kshefchek, line 232: This class is forked from the bbop-widgets library, ideally we should try to achieve this in a different portion of the code (handler.js ?), or feedback to @kltm on how to make the code more modular.  My hope was that we revert back to the code in the bbop library at some point and remove this from the monarch code base.

kltm: Well, it depends on what you're trying to do here. References to actual fields should not exist in the widget, so it would need to be: hinted out in the metadata YAMLs (preferred), single field override in the handlers data (somewhat uglier), or passed as a special-purpose parameter (ick).

kltm: Honestly, if I'm understanding this right, it would be two separate YAMLs--one for each pseudo-view.

kshefchek:
+1

templates/node.mustache
```diff
@@ -0,0 +1,152 @@
...
+                    </div>
+                    {{#hasBrw}}
+                    <div class="row">
+                        <br>
+                        <div class="col-xs-12">
+                            <div class="panel panel-default">
+                                <div class="panel-heading" style="padding:5px 10px 7px 10px;">
+                                  <h5 class="panel-title">{{node_label}} Classification
+                                    <a
+                                      class="pull-right"
```
jmcmurry, line 116: @DoctorBud if we do class = "pull-left" will it align the info circle to just after `Phenotype/Disease Classification`? If so, that would be great as it is currently so inconspicuous hard to see it is there, especially when browser window occupies whole screen width.

jmcmurry: Hi @DoctorBud, did you see this comment? The info circle is so far to the right that it is hard to visually associate it with the text. Especially when browser is at full width of screen.

screen shot 2016-10-20 at 11 35 25 am

PR #1360, opened 2016-09-15
Point to GitHub for recent improvements/issues

jmcmurry:
Until such a day as we are able to systematically communicate release notes, let's at least not look moribund.


templates/footer.mustache
```diff
@@ -2,21 +2,17 @@
...
     {{/monarch_footer_fade_p}}>
-    The Monarch Initiative is a collaboration between Oregon Health & Science University,
-    University of California, San Diego, Lawrence Berkeley National Laboratory, the University of
-    Pittsburgh, Sanger Institute, Charité - Universitätsmedizin Berlin, and Garvan Institute of Medical Research
-    and is supported generously by a NIH Office of the Director Grant #5R24OD011883.<br/>
-    Except where otherwise noted, this work is licensed under a Creative Commons Attribution
-    3.0 License.
+    The Monarch Initiative is a collaboration between Oregon Health & Science University, The Jackson Laboratory, Lawrence Berkeley National Laboratory, the University of
+    Pittsburgh, Sanger Institute, Charité - Universitätsmedizin Berlin, Garvan Institute of Medical Research, and William Harvey Research Institute, Barts & The London School of Medicine & Dentistry, Queen Mary University of London.
+    Monarch is supported generously by a NIH Office of the Director Grant #5R24OD011883, as well as by NIH-UDP: HHSN268201300036C, HHSN268201400093P, NCI/Leidos #15X143.<br/>
```
DoctorBud: These source lines are annoyingly long. In the future, please use a 80-100 character word wrap text editor.

templates/page/releases.mustache
```diff
@@ -6,6 +6,16 @@
     <div class="content content-bottom">
       <div class="title">Monarch Releases</div>
 	      <div class="section">
+	      <div class="sectiontitle">Fall 2016</div>
+		    <div class="text">The Monarch interface, data, ontologies, and API are rapidly evolving. 
+		    Recently closed issues can be found at <A HREF="http://bit.ly/monarchclosedissues">http://bit.ly/monarchclosedissues</a>
+		    Open issues can be found at <A HREF="http://bit.ly/monarchissues">http://bit.ly/monarchissues</a>. 
+		    As always, we value your feedback. If you have any problems, concerns, or 
+		    suggestions, please 
+		    <A HREF="http://bit.ly/monarchnewissue">log a new GitHub issue</a>,
```
DoctorBud: HREF should be lowercase, as should the A tag.

DoctorBud:
+1 with some comments about future similar changes.

templates/page/releases.mustache
jmcmurry: Not sure why it was capped in the first place, but just kept it consistent.

templates/footer.mustache
jmcmurry: done

jmcmurry:
@mellybelly, @kltm, @cmungall, @DoctorBud are you opposed to the idea of bitlys to track engagement? They are clearly _not_ ideal from perspective of user trust, but on the upside, they're memorable, trackable, short, free, and easy. 

There is no shortage of much better ways to track engagement, but we don't have the bandwidth to implement any of them right now.

I don't feel strongly about using bitly in lieu of links that are easy to find eg. https://github.com/monarch-initiative/monarch-app/issues; however, it is especially handy for things like the cross-repo issues list. 
https://github.com/issues?q=is%3Aissue+is%3Aopen+user%3Amonarch-initiative+sort%3Aupdated-desc

Links such as these require some more esoteric knowledge about the github API and are pretty hard to get back to if you start clicking around GitHub.

Maybe I'm just to curious for my own good and the bit.ly are unnecessary; after all it is not horrible to just use clean anchor text with a non-redirected hyperlink. It is just that it throws away another free opportunity to figure out what in the world people are doing on our site and what they want. Perfect is the enemy of the good. It is also the enemy of the perfect.

kltm:
@jmcmurry Engagement for what exactly? If you are generally wanting to use bit.ly at various points in the app to track people, that would not be good. (Privacy, management, granularity, etc.)

kltm:
As well, shorteners are a a sign that something has gone wrong. It's one thing to text a friend a shortened URL, it's quite something else to see them being used in earnest. As well for the users, they have _no_ idea what is at the other end until they click, which is worrying for security, etc.

jmcmurry:
Just for the record, _never_ wanted to use them "in earnest" throughout the site. Just in places like this static page where we want to know to what degree users have feedback but stop short of actually submitting it, for instance. That would suggest we need to lower the bar somehow. Or how interested users are in our recent development. There are enough reasons to take them out, so I did.

kltm:
Maybe we can circle back around on analytics again at some point. What case are you looking at where you're not getting what you want picked up by GA?
GA, is some modes, also supports link tracking, but I think given the nature and possible sensitivity of the data on the site, it might we worthwhile to look at a private installation of piwik (or something) to keep things centralized.

jmcmurry:
@kltm I would love that, but also aware it is yet one more competing priority. 
Let's get search improvements wrapped up and then talk about it.
Thanks so much.

iamkoehler:
I also use GA on our sites and think it might be the better option. What stops you from using it?

jmcmurry:
We already use GA but not really to its full advantage.

PR #1362, opened 2016-09-15
Team updates: Tudor, Craig, Edwin, Tom, Damian, Nicole

jmcmurry:


kshefchek:
Let's also add Tom while we're at it.

PR #1367, opened 2016-09-20
Decruft (don't merge yet)

jmcmurry:
@kshefchek please have a look at /tests/urltester.js to see what can be deleted, perhaps whole file?


kshefchek:
Some of the tests in the urltest class seem sane, but this hasn't been part of our testing process for some time now.  I think we could update them or clear them out, @cmungall I imagine you wrote these?

cmungall:
yes, they can be eliminated

PR #1368, opened 2016-09-21
How to cite monarch (don't merge yet)

jmcmurry:
@mellybelly @cmungall please review. I know it isn't perfect, but we need something there.

### DIFF
--- a/templates/page/citation.mustache
+++ b/templates/page/citation.mustache
@@ -15,25 +15,67 @@
         Melissa A. Haendel GENETICS August 1, 2016 vol. 203 no. 4 1491-1495; DOI: 10.1534/genetics.116.188870
 		    </div>
 		    <div class="text">
+		     
+		    </div>
+		    <div class="text">
 		    MonarchInitiative.org: Integrated data and analyses connecting phenotypes to genotypes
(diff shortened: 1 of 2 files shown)


mellybelly:
+1
The only thing I might add is a line relating to when the base URL is cited, eg. www.monarchinitiative.org, we prefer that the access date be included. Ideally we'd version Monarch data, but this hasn't happened yet, and until then, access date is best I would think.

jmcmurry:
I'm not sure I understand you; access date is already part of the recommended citation for the data pages.
Also, under which circumstances would you want the base URL to be cited instead of the paper? What do you want to use as the database identifier of record for Monarch? 
- RRID
  - RRID:nlx_152525
  - RRID:SCR_001373
- NAR
  - Still in process
- Re3data
  - r3d100011594

mellybelly:
hmm apologies I missed that, I see it now, e.g.
Monarch Consortium. Accessed 2016-09-22. Marfan Syndrome. https://monarchinitiative.org/OMIM:154700

that looks good. I don't think any of the above style db identifiers are required at this time. If we really want one of those, I would register as biodbcore as per NAR/ISB recommendations, for example: https://biosharing.org/biodbcore-000007 
Biosharing has promised resolution and persistence but I don't think this is released yet.

DoctorBud:
I'm going to close this PR. It's been hanging here for two years doing nothing and is probably moot. If this issue is explored again, a new PR should be created or this can be resurrected if the conflicting files are updated.

Issue #1239, opened 2016-04-21
Errors reported by GrayLog


jnguyenx:
This is a dump of the reported log errors. I don't think that they're all relevant, some should be silent.

```
Debug: internal, implementation, error 
    TypeError: Uncaught error: Cannot read property 'Content-Disposition' of undefined
    at wrapResponse (/opt/monarch-app/lib/monarch/web/webenv.js:289:35)
    at /opt/monarch-app/lib/monarch/web/webenv.js:385:25
    at /opt/monarch-app/node_modules/wait.for/waitfor.js:15:31
```

```
errorResponse:Cannot find apple-touch-icon (404)
Stack trace=Error
    at errorResponse (/opt/monarch-app/lib/monarch/web/webapp.js:775:19)
    at notFoundResponse (/opt/monarch-app/lib/monarch/web/webapp.js:794:12)
    at Object.resolveByIdHandler (/opt/monarch-app/lib/monarch/web/webapp.js:5353:22)
    at /opt/monarch-app/lib/monarch/web/webenv.js:384:52
    at /opt/monarch-app/node_modules/wait.for/waitfor.js:15:31
```

```
errorResponse:Cannot find apple-touch-icon-precomposed (404)
Stack trace=Error
    at errorResponse (/opt/monarch-app/lib/monarch/web/webapp.js:775:19)
    at notFoundResponse (/opt/monarch-app/lib/monarch/web/webapp.js:794:12)
    at Object.resolveByIdHandler (/opt/monarch-app/lib/monarch/web/webapp.js:5353:22)
    at /opt/monarch-app/lib/monarch/web/webenv.js:384:52
    at /opt/monarch-app/node_modules/wait.for/waitfor.js:15:31
```

```
Debug: internal, implementation, error 
    Error: Uncaught error: ENOENT: no such file or directory, open './dist/app.26a4249946448350be6b.css'
    at Error (native)
    at Object.fs.openSync (fs.js:549:18)
    at Object.fs.readFileSync (fs.js:397:15)
    at Object.fs_readFileSync (/opt/monarch-app/lib/monarch/serverenv.js:23:37)
    at Object.wrapContent (/opt/monarch-app/lib/monarch/web/webenv.js:358:24)
    at Object.<anonymous> (/opt/monarch-app/lib/monarch/web/webapp.js:1178:24)
    at /opt/monarch-app/lib/monarch/web/webenv.js:384:52
[...]

PR #1374, opened 2016-10-06
Reduces the amount of extraneous logging when invalid paths are specified.

DoctorBud:
This is mostly in response to https://github.com/monarch-initiative/monarch-app/issues/1239
- Reduces the amount of extraneous logging when invalid paths are specified. For example, /apple-touch-icon-precomposed or /docs/foo. In most cases, these invalid paths were due to bots or cached older versions of the webapp, or due to Apple-specific browser behavior (cf http://stackoverflow.com/questions/5110776/apple-touch-icon-for-websites/21144916#21144916).
- Eliminates a round-trip to SciGraph when an invalid path is specified. For example, /foo used to ping SciGraph, just in case foo is an id.
- Ensures that the webenv.wrapContent() method checks for file existence prior to trying to read it. This eliminates an annoying stack crawl in the log files.
- Fixes the package.json 'run start' script so that it uses 'npm run build' rather than just 'gulp build'.

### DIFF
--- a/lib/monarch/web/webapp.js
+++ b/lib/monarch/web/webapp.js
@@ -428,7 +428,7 @@ function buildEngine(defaultConfig, golrConfig) {
     },
     store: function(tbl, key, val) {
         var path = "./cache/"+tbl+"/key-"+key+".json";
-        console.log('###CACHE SAVE:', key);
+        // console.log('###CACHE SAVE:', key);
         env.fs_writeFileSync(path, JSON.stringify(val));
     },
     clear: function(match) {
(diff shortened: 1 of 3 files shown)


jnguyenx:
Looks good! Just one comment: all the non found paths will be silent now? It might still be good to get an output error for those, we can have broken links within our app.

DoctorBud:
@jnguyenx I didn't silence bogus paths, but I did ensure that only one line (OK, sometimes 2 lines) will be emitted to the log for each bogus path, and that a correct Error page will be generated. This is an improvement over the stack crawls and multiple messages we were getting. For example, the logs look like this after using `/bogus1`, `/docs/bogus2`, and `/bogus3/bogus4/bogus5`. 

```
errorResponse:No such path: bogus1 (404)
###wrapContent( ./docs/bogus2 ) does not exist
errorResponse:Path not found: ./docs/bogus2 (404)
errorResponse:Page Not Found: /bogus3/bogus4/bogus5 (404)
```

jnguyenx:
Excellent! Then it looks good to me for merging.

kshefchek:
+1 as well

PR #1379, opened 2016-10-21
facet with checkboxes rather than clicking on legend labels for barcharts

kshefchek:
Utilizes checkboxes instead of requiring a user to click on the legend key itself for faceting on species for our barcharts.

Note I iterated on this pretty quickly for the NAR re-submission and it is mostly untested.


kshefchek:
This is ready to be merged.  Note that in order to have a checkbox form within the svg itself, we need to use a foreignObject tag and there may be issues with older browsers.  @yuanzhou @kltm curious if you know of good workarounds. There's also some oddities with the checkboxes not resizing when zooming.

kltm:
Back when we did a lot more SVG stuff (pre-D3, etc.) we used various toolkits; the last iteration was: "net.carto.checkboxAndRadiobutton" (http://www.carto.net/papers/svg/gui/checkbox_and_radiobutton/), which worked really well, but is now defunct (?).
From our experience though, doing raw GUI stuff was really harry; maybe this has improved with foreignObject.

kltm:
Although, I would add that since SVG can directly take standard JS events and the like, things like checkboxes are pretty easy to implement "raw".

kshefchek:
Implement raw as in making the checkbox with svg elements - rect, path, etc.?  I considered that as well.

kltm:
Nested rect, that listens for event and changes color. Not as nice as being able to probe a form with jQuery or something, but "easily" doable.

Also, there are other abstractions out there now, like:

http://jsxgraph.uni-bayreuth.de/docs/symbols/Checkbox.html
http://snapsvg.io/docs/#Element.click

(From: https://developer.mozilla.org/en-US/docs/Web/SVG/Tutorial/Tools_for_SVG)
But I'm not sure how well that would play with what you're using now...

kshefchek:
That's similar to what we have on production, a user clicks on the legend text or rect and the opacity changes, but this is confusing, but we could easily add a second rectangle for the sole purpose of having a checkbox function that might be more obvious.

Issue #1386, opened 2016-10-25
High priority modifications to front-end of search results


jmcmurry:
Of course dependent on #1383, and pursuant from #1317, below is a distillation of high priority improvements in order of importance:
## Faceting by result type:
- [x] Gene*
- [x] Disease\* 
- [x] Phenotype

`*` Note that while disease GROUP and gene GROUP are useful distinctions for each result, they need not be in facets as this may lead to confusion. Other facets will be considered in future.
## Faceting by species:
Species facets should be common names wherever possible and ordered by frequency of search, eg.
- [x] human
- [x] mouse
- [x] fish
- [x] fly
- [x] worm
- [x] dog
- [x] cow
- [x] cat
- [x] pig
- [x] chicken
- [x] other
## Facet styling
- [x] Facets must be on the left of the page and vertically oriented
- [ ] Facet counts must dynamically update with counts (from all results not just first page)
  Note that because both sets of categories are disjoint, it is not necessary to support ternary [+/-/either] styling. Checkboxes are adequate and even without checkboxes is fine as a first pass.
## Results styling
- [x] Intro to results currently says "Results for "xxx" searching phenotypes, diseases, genes, and models", however we are not indexing models yet and should remove this in meantime.
- [x] Highlighted string must be shown in bold, this includes support for portions of words
- [ ] Gene results must include both the symbol and the name (Concatenated into the same column is fine for simplicity, else the matrix gets awfully sparse when interdigitated types come into play eg. "LHX 1 (LIM homeobox 1)"
- [ ] Taxon for diseases is named (99.99 % will be human, but needs to be stated)
- [ ] Taxon for phenotypes is named (eg. human, mouse, fly etc. Note that some phenotypes eg. GO cellular processes may require special handling) 
- [ ] Once a category filter is selected, remove that column from the results table
- [ ] On mouseover, highlight the active row and make any part of it clickable
- [ ] Update summary text (or representation) to reinforce which facets are selected: 
<img width="669" alt="screen shot 2016-11-03 at 3 29 51 pm" src="https://cloud.githubusercontent.com/assets/3658863/19987741/f49773f6-a1d9-11e6-8dad-3b0ad806808a.png">

## Results paging
- [x] State total number of results: eg 1-50 of 450 results for **"LHX"**
- [x] Enable user to load more results
## Search box
- [x] Don't blank out the search box, keep the searched-for-term in there so the person doesn't have to retype it to refine it.

[...]

PR #1388, opened 2016-11-03
[PLEASE MERGE] Site search redesign

yuanzhou:
As a result of the #1386, this PR redesigned the whole site search using the new golr api. Facets are rendered based on the golr results and users can use these filters to narrow down the results.

capture

When clicking the "Load more" button, new data is loaded thanks to the pagination functionality in the new search api.

6666666re

@jnguyenx Please take a look and review the code as well as UI. As you mentioned earlier, once the `highlight_documents()` is tweaked, I'll switch back to it.


yuanzhou:
And the missing taxon is considered as a data issue that needs to be addressed later, so the filters can cover all results.

jmcmurry:
@kltm @jnguyenx @yuanzhou I get that data should drive the facets (and agree), however:

1)  "sequence feature" is a superset of genes and other stuff (variants, genotypes etc) none of which are currently indexed. Thus, I would like to silence this category until we get our house in order.

2) When we do get our house in order, I would like the category hierarchy to be clear in the UI. Is this hierarchy part of the data model as returned?

3) It would still be useful in the UI to do some post-processing to make, for example, the order of the taxa to prioritize the species of typically greatest interest, at a _minimum_ prioritized within the facet list. How do you propose we do this and still have a data driven display?

yuanzhou:
@jmcmurry 

1) previously I excluded "Class" and "sequence feature" from the category list, @kltm and @jnguyenx suggested to do everything as data-driven. I can still exclude these two categories, but maybe I'll also need to exclude them from the table rows. If this can be done on the data end or Golr, then it's gonna be less client code.

3) Right now I sorted the taxa in alphabetic order.

jmcmurry:
Thanks @yuanzhou :) 

As the groupings are in every case species (not order/phyla etc), I think it is fine to just say "All species" (side note, plural of 'taxon' is 'taxa'). 
Also, why is there a discrepancy between class count and total result count?

yuanzhou:
@jmcmurry Just made the changes. Thanks for correcting my typo.

I don't see a discrepancy between class count and total result count

jmcmurry:
@mellybelly says human, mouse, zebrafish, fly, worm, and then everything else by data richness

yuanzhou:
Based on the meeting discussion, I'll make the following enhancements:

1) Sort species: human, mouse, zebrafish, fly, worm, and then everything else by data richness

2) Make the filters (on the left side) and count info as well as table header stay when scrolling.

3) Show the current search term in the search input field instead of using the default place holder

I'll notify @jnguyenx once it's ready to be merged.

jnguyenx:
@yuanzhou I don't think that hiding Class and sequence feature in the UI is the right call, I'll change in solr.

You must be sick of me saying the same thing, but really let the data drives the UI. Can you please revert this change?

yuanzhou:
@jnguyenx that's what I really wanted to hear :) I can revert the changes very easily once you have the solr results tweaked. Please also make sure to hide the "Class" and "sequence feature" from each individual `doc.category`. 

One more tiny thing is to make "Phenotype" in both `facets_fileds` and `doc.category` lowercase for consistency with "gene" and "disease". Right now I had to convert it into lowercase in order to render the hyperlink correctly.

And are you gonna also sort the species in solr or this is something I should do on my end?

jmcmurry:
@yuanzhou once this is merged feel free to check off the completed tasks within https://github.com/monarch-initiative/monarch-app/issues/1386

jnguyenx:
I think that the Phenotype vs phenotype is a bigger change: https://github.com/monarch-initiative/configs/issues/23
Let's see first how that evolves, in the meantime you can keep your workaround.

Solr can sort facets only by count (what we have currently) or alphabetically. The way @jmcmurry wants it has to be done manually on the client side.

jmcmurry:
@jnguyenx agreed; that kind of sorting is not appropriate for solr anyway.

jnguyenx:
@yuanzhou I filtered out some categories on solr.

yuanzhou:
@jnguyenx thanks for excluding "Class" and "sequence feature" on solr, I've also removed my check of them on the client side. Now I think we are good to go with merging this PR. Once it's deployed on beta, we can do more tests and fix/improve things.

lib/monarch/web/webapp.js
```diff
@@ -194,6 +198,7 @@ var pup_tent_dirs = [
                     'node_modules/d3',
                     'node_modules/font-awesome/css',
                     'node_modules/font-awesome/fonts',
+                    'node_modules/isotope-layout/dist', // used by search results - Zhou
```
jnguyenx: I don't think you're currently using this.

lib/monarch/web/webapp.js
```diff
@@ -1055,8 +1060,136 @@ web.wrapRouteGet(app, '/node_modules/phenogrid/:filename', '/node_modules/phenog
...
+
+    try {
+        // Only display 30 results per page start from row 1, index 0
+        // This can be used for paging
+        var resultsCountPerPage = 25;
+        //var results = engine.autocomplete(term, null, resultsCount);
+        
+        // TODO new implementation for Zhou
+        // Remember to convert JSON string to JSON object
+        var golrResults = engine.search(term, JSON.parse(filters), ["category", "leaf", "taxon_label"], resultsCountPerPage, (pageNum - 1)*resultsCountPerPage);
```
jnguyenx: You don't seem to use the leaf faceting in the UI, you should remove it from the query.

lib/monarch/web/webapp.js
jnguyenx: Actually it makes the build fail locally.

yuanzhou: You are right, forgot to remove that line. Do it now. Thanks!

lib/monarch/web/webapp.js
yuanzhou: Got rid of leaf faceting from the call. Thanks!

lib/monarch/web/webapp.js
```diff
@@ -1055,8 +1059,136 @@ web.wrapRouteGet(app, '/node_modules/phenogrid/:filename', '/node_modules/phenog
...
+        // Default to 0
+        var count = 0;
+
+        var resultsOutput = '';
+
+        // Default to true to show the load more button
+        var loadMoreBtn = true;
+
+        // object with rendering functions
+        if (golrResults.response.docs.length > 0) {
```
jnguyenx: Instead of getting the from the raw json, you can go through the bbop_response_golr API, that'll be a better abstraction I think.

lib/monarch/web/webapp.js
```diff
@@ -1055,8 +1059,136 @@ web.wrapRouteGet(app, '/node_modules/phenogrid/:filename', '/node_modules/phenog
...
+            count = golrResults.response.numFound;
```
jnguyenx: Same as above.

lib/monarch/web/webapp.js
```diff
@@ -1081,56 +1213,189 @@ function searchHandler(request, term) {
...
-        var resultsCount = 50;
-        var results = engine.autocomplete(term, null, resultsCount);
+        // Only display 30 results per page start from row 1, index 0
+        // This can be used for paging
+        var resultsCountPerPage = 25;
 
-        // TODO new implementation for Zhou
-        //var results = engine.search(term, {}, ["category", "leaf", "taxon_label"], resultsCount, 0);
+        // Only faceting with category and taxon_label - Zhou
+        var golrResults = engine.search(term, {}, ["category", "taxon_label"], resultsCountPerPage, 0);
```
jnguyenx, line 1222: I don't really like the data workflow that we have, if I'm not mistaken we get the first page from the /search handler and the rest of the pages through the load button which calls /searchfiltering.
In my opinion it should be consolidated in the way that all the data should come from /searchfiltering. The initial call to /searchfiltering can be triggered by the client with default values.

This means major refactoring on the following code, so I understand if you don't want to change the current implementation.

kshefchek:
@yuanzhou could you pull from master?  I fixed some failing behave tests.

yuanzhou:
@kshefchek thanks for fixing the behave tests. I've also fixed the broken search test. And later on we can add more search tests since we may have some more tweaks after this PR gets merged.

jnguyenx:
@yuanzhou so for me the only nice thing fixable rapidly is to use the bbop-golr-response everywhere instead of probing the json object manually.

Otherwise looks good for merging!

yuanzhou:
@jnguyenx I looked through bbop-response-golr code and didn't seem to understand where in the webapp.js search endpoint I should call bbop-response-golr functions other than just using `documents()` to get the highlighted docs. Please advise.

lib/monarch/web/webapp.js
jnguyenx: Here for example use documents() instead of response.docs.

lib/monarch/web/webapp.js
jnguyenx: Here you can use total_documents() instead.

jnguyenx:
I added some examples, ultimately everything you can use [from here](https://github.com/berkeleybop/bbop-response-golr/blob/master/lib/response.js) is a win.

yuanzhou:
@jnguyenx where can I find those newly added examples? I checked out the tests but didn't see too much. And the documentations doesn't have concrete examples. I really don't see the point of using bbop-response-golr to handle simple JSON rendering. Maybe we can just merge this PR then you can update the code with bbop-response-golr functions.

jnguyenx:
Okay I don't know how to use the github core review tool... :disappointed: 

Now you should be able to see my comments.

lib/monarch/web/webapp.js
yuanzhou: The `total_documents()` at https://github.com/berkeleybop/bbop-response-golr/blob/master/lib/response.js#L222 simply returns the `robj.response.numFound`, which is the total count instead of the count of this current page. `golrResults.response.docs.length` returns the number of results of the current load.

lib/monarch/web/webapp.js
yuanzhou: The `total_documents()` at https://github.com/berkeleybop/bbop-response-golr/blob/master/lib/response.js#L222 simply returns the `robj.response.numFound`, which is the total count instead of the count of this current page. `golrResults.response.docs.length` returns the number of results of the current load.

lib/monarch/web/webapp.js
yuanzhou: Yes, the /search returns the first page of the results as well as the facets that are rendered on the left side. /searchfiltering is triggered when clicking the Load more button or the filters of category/taxon. So basically the /searchfiltering handles actual paging. This workflow works pretty well right now, maybe we can improve it in the next round and add more features.

lib/monarch/web/webapp.js
jnguyenx: Yep exactly, so instead of `count = golrResults.response.numFound;` you can use `total_documents()` and instead of `golrResults.response.docs.length` you can use `documents().length` if I'm not mistaken.

lib/monarch/web/webapp.js
yuanzhou: Another reason I separated "/searchfiltering" from "/search":

The search URI route is: "/search/:term"

The search filtering URI route is: "/searchfiltering/:term/:filters/:pageNum"

If you want to merge them into one route, then "filters" and "pageNum" parameters will have to be required parameters. Then the default search page will be rendered with default filters and pageNum in the URL, which is not good.

yuanzhou:
@jmcmurry any reason that this PR shouldn't be merged? Further improvements can be discussed and implemented in the next round.

capture

lib/monarch/web/webapp.js
jnguyenx: Not sure to understand why it causes an issue, but let's revisit that later. We discussed using url parameters last time for filters and paging, I'm curious why you chose to use url paths instead.

yuanzhou: Because I don't have too much knowledge of how those routes work in webapp.js. I just followed some other existing route examples and they are using paths. I can look into that later and combine the /search and /searchfiltering into one call.

PR #1393, opened 2016-11-10
Bumped phenogrid version to 1.3.10

yuanzhou:
@kshefchek feel free to merge if all checks pass.


kshefchek:
@yuanzhou was I supposed to update the files in the dist directory or is gulp ran automatically when installing?

yuanzhou:
@kshefchek you are right, I also forgot to create the new bundled files in the /dist folder. Just did it with `gulp bundle` and bumped to Phenogrid 1.3.10.

PR #1399, opened 2017-01-03
switch back to prod services

kshefchek:
switch back to prod on beta to fix failing tests.


kshefchek:
We're failing now due to:
Error: certificate has expired

Looks like we need to update our ssl cert, cc @jnguyenx

PR #1407, opened 2017-02-02
Removing step 1

cmungall:
Removed step 1 as this is now cruftier than a professor of cruftiness at cruft university


jmcmurry:
Yeah; that was written back when we were all still transitioning to the new workflow (and had nonforked branches of master). We can safely remove it now.

DoctorBud (approved):

PR #1418, opened 2017-02-24
[#1386] make filters update on change

jnguyenx:
Unfortunately I couldn't do what I wanted due to the current state of the code. There was some work pending which has never been taken care of:
https://github.com/monarch-initiative/monarch-app/pull/1388#discussion_r87058149

This is a temporary fix, as I think that we'll need to rewrite this piece of code to have something more flexible and simpler.

Note that I removed the taxon ordering as it breaks the dogma to let the data drives the UI.

### DIFF
--- a/js/search_results.js
+++ b/js/search_results.js
@@ -39,6 +39,7 @@ $( document ).ready(function() {
         }
       }
 
+      // TODO change that to a POST request (Jeremy)
       // Convert object to JSON string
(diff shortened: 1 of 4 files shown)


kshefchek:
@jnguyenx could you retest with our prod services?

kshefchek (approved):

jnguyenx:
Let's do things properly and proceed with the rewrite. I've been mending things up for the Monarch All Hands meeting, as a temporary solution.

PR #1427, opened 2017-03-07
READY: Fixing up technical documentation on main site

cmungall:
See https://github.com/monarch-initiative/monarch-app/projects/2


kshefchek:
It could be useful to bring some of these docs back, especially the ones on the annotation sufficiency meter.  Maybe a candidate for a new markdown doc workflow.

Issue #420, opened 2014-06-07
clean up repeated disease-gene rows for equiv ids


nlwashington:
on a page like:

http://localhost:8080/disease/OMIM_104300

the gene tab lists many disease-gene associations, and some seem like they might be different diseases, but they aren't.

for example, we see:
OMIM:104300 associated with APP
and we also see:
MESH:D0000544 associated with APP

however, the ontology says that these two ids are equivalent.

these should probably be collapsed into a single row.
this will probably require reconciling each item in the association table with the equivalence axioms (though they probably are already pulled into the page from the previous ontoquest call).

nlwashington:
@kshefchek changed things up; now that this is working off of scigraph ontologies, any equivalent ids are grouped together.  This is as good as it's going to get for now, until KEGG ids get into the disease ontology.  Leaving open as a reminder.

jmcmurry:
This is a duplicate, and overtaken by events.

PR #1437, opened 2017-03-13
switch beta to dev services

kshefchek:
Testing dev servers.

### DIFF
--- a/conf/server_config_beta.json
+++ b/conf/server_config_beta.json
@@ -3,16 +3,15 @@
 
     "app_base" : "https://beta.monarchinitiative.org",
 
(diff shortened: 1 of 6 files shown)


kshefchek:
We need to make some minor testing changes with the new search index.  Also we lost https://www.ncbi.nlm.nih.gov/clinvar/variation/159648/, @TomConlin any reason we might be filtering this out?  Will change the test so it passes.

TomConlin:
> We need to make some minor testing changes with the new search index. Also we lost
> https://www.ncbi.nlm.nih.gov/clinvar/variation/159648/, @TomConlin any reason we might
> be filtering this out? Will change the test so it passes.

It would have to be related to #420 
I do not see SCV000194552 in the reject pile either.  checking deeper

PR #1493, opened 2017-09-19
Disable phantomjs. Add progress bar. Fix Source width. Fix unit test

DoctorBud:
- Disable phantomjs install in travis.yml and disable phantomjs tests.
- Add progress bar to search window.
- Add v-clock directive to search window so that raw HTML is not displayed on load.
- Fix width of the Source column in the various tables.
- Adjust maui_autocomplete to work with new data.


kshefchek:
+1

PR #1499, opened 2017-09-28
update orgs on footer

lwinfree:


lwinfree:
@kshefchek my checks are failing on this one too - same checks failing as my other PR. Please let me know if I need to change anything/can help with the failing checks!

DoctorBud:
Did you test the app to see how this footer looks? I'm concerned it will make the footer quite large vertically and that we should verify that it doesn't affect pages with footers. IMHO, a footer should be less than 10% of the 'visible' page height if we are putting it on every page. Otherwise, it's just annoying.

lwinfree:
@DoctorBud Yeah I checked it out & got a second opinion from @jmcmurry. I agree with you that it seems pretty big...I'm open to ideas for making it smaller.

jmcmurry:
Given how many institutions are involved (and growing), I'm open to linking to the team page rather than listing all.

DoctorBud:
My latest PR, https://github.com/monarch-initiative/monarch-app/pull/1515, included these changes to footer.mustache, so this particular PR is no longer necessary. I am closing it.

PR #1508, opened 2017-10-10
sample myvariant widget - do not merge

kshefchek:
Demo myvariant widget, cc @putmantime


kshefchek:
This is failing for a test that checks that both genes and variants show up on disease-variant tables, specifically https://beta.monarchinitiative.org/disease/OMIM:182212#variants

The issue is for ClinVarVariant:409977, it should connect to NCBIGene:6497. @TomConlin could you take a look?  I see all the variant, disease, gene parts for the RCV in the xml file, but I'm not aware of all the parsing logic.

It doesn't seem to be an issue across all variants so maybe we can temporarily disable or change the ID.

kshefchek:
I think the build is failing now because our beta owlsim is on the fritz from the pub experiment, that should be fixed soon.

PR #1509, opened 2017-10-13
add contact request

lwinfree:
to links page - add sentence about emailing us if they use our info/link us as per @jmcmurry request


DoctorBud:
+1

PR #1515, opened 2017-10-31
CSS Simplification/Deletion. Style unification/consistency. Bug Fixes. Footer fixes.

DoctorBud:
The overall goal of this PR is to reduce the amount of custom CSS used for styling, and instead to rely upon the Bootstrap default behavior. When we do need custom styles, I updated the CSS to explicitly scope these styles so we don't have general styles like `.title` polluting the CSS namespace. This will make it easier to ensure uniform styling throughout the app, as well as making it easier to change that styling via a 'theme' stylesheet. Along the way, I fixed some bugs and learned a bit more about the challenges of building a SPA (single-page application) upon the Monarch codebase.

- Clean up CSS to avoid collisions during bundling.
- Bootswatch theme overrides now working correctly.
- jquery-ui and phenogrid are being managed and bundled better.
- No more LESS. monarch.scss is now the root stylesheet.
- Lots of CSS simplification, elimination, unification
- Makes the About and Documentation pages more uniform in their styling
- Fixed Data Sources table to not exceed page width
- Integrated Vue into the bundle and adjusted search/exac to work with that.
- Fix `http` link to be `https` in README.md
- Fix `Analyze` button so it is centered.
- Add bootswatch.com themes to allow for easy switching between themes to help debug CSS issues.
- Delete `angular-chromosome-vis.js` and `angular-chromosome-vis.css` which are not used.
- Use and bundle phenogrid's CSS directly rather than relying upon it's bundled version.
- Fix bug in GenomeFeatureViewer where `externalURL` was used but not defined.
- Fix search_results.js to be a CommonJS module rather than a jQuery onload fragment
- Added a `/page/styles` route to make it easier to see the effect of CSS changes (for debugging and not for final production).
- In `api.js`, fix default owlsim URL to use `https` (most likely, we never use the default, so this is probably moot).
- Eliminate pup-tent inclusion of Bootstrap and jQuery, as they are in the Bundle already.
- Use `https` where we still refer to `http`, since all of our services now have https.
- Bundle phenogrid rather than using pup-tent.
- Fix  bug in `widgets.js` where an HTML table was being generated without a closing tag due to a typo.
- Delete chromosome-vis-demo /labs example
- Fix Documentation/Phenogrid page so that the title isn't buried under the navbar.
- Updated code docs in a few places.
- Updated footer as per Lilly's request.
- Deleted unused CSS files


kshefchek:
+1

PR #1521, opened 2017-11-17
Vue-router and a Vue SPA integrated with Monarch. CSS and bug fixes.

DoctorBud:
vue-router replaces Navigo Router.
Vue-router and a Vue SPA integrated with Monarch.
Autocomplete looks better.
SPA mode almost complete using vue-router.
Adds `/image/person-unknown.png` for missing images on Team page.
Restores some of the CSS that was aggressively deleted in previous PR.
Non-SPA and SPA CSS are more closely matched.
Fixes various `.mustache` templates so that their HTML is properly nested. VueJS was flagging these as errors.
Adds Autocomplete Search form to SPA Home Page (this might be enabled in non-SPA mode later).
Adds `vue-loader` to webpack config to support single-file VueJS components.


kshefchek:
The footer looks a little odd when there is not content covering the length of the page, for example: /gene/NCBIGene:388552#diseases

pr

In general it would be nice to restyle the analyze phenotypes page - I prefer the styling up on production but understand the css clean up.  Maybe we could re-design this widget all together.

DoctorBud:
@kshefchek I improved the Footer and the Analyze Phenotypes styling. Please check them. Your feedback is useful.

kshefchek:
+1

PR #1531, opened 2017-12-18
Unit tests working

DoctorBud:
- Fixes behave tests to deal with MONDO changes.
- Eliminates use of implicit_wait in behave tests, which may increase the speed of testing.
- Enables behave tests to work properly in both SPA and normal mode.
- Eliminates the CSS class content-bottom.
- Fixes API tests and behave tests to dodge the problem where non-MONDO aliases do not have the same information as their MONDO root.


kshefchek:
+1

PR #1536, opened 2018-01-31
- Adds Node.vue component.

DoctorBud:
- Adds Node.vue component.
- DRYs up webapp.js a little.
- Navigator works properly with overlay.
- Layout adjustment.
- Fixes double-fetch except when /resolve is used (TBD).
- Integrate Patternfly SCSS
- Supports 'npm run build' version of SPA
- Ensure that the USE_SPA correctly switches between Bootstrap and PatternFly modes.
- Adds NodeCard component


kshefchek:
+1

PR #1541, opened 2018-02-15
Add table component

putmantime:
Works for disease to phenotype associations

### DIFF
--- a/ui/components/Node.vue
+++ b/ui/components/Node.vue
@@ -127,23 +127,7 @@
         v-if="expandedCard"
         class="expanded-card-view">
         <h3 class="text-center">{{expandedCard}} Associations</h3>
-        <table
-          class="fake-table-view">
-          <thead>
-            <th v-for="colName in Array.from('ABCDEFGH')">{{colName}}</th>
-          </thead>
-          <tbody>
-            <tr v-for="rowNumber in Array.from('0123456789')">
-              <td v-for="colData in Array.from('ABCDEFGH')">({{rowNumber}}, {{colData}})</td>
-            </tr>
-            <tr v-for="rowNumber in Array.from('0123456789')">
-              <td v-for="colData in Array.from('ABCDEFGH')">({{rowNumber}}, {{colData}})</td>
-            </tr>
-            <tr v-for="rowNumber in Array.from('0123456789')">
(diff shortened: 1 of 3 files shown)


ui/components/Node.vue
```diff
@@ -127,23 +127,7 @@
...
-            </tr>
-            <tr v-for="rowNumber in Array.from('0123456789')">
-              <td v-for="colData in Array.from('ABCDEFGH')">({{rowNumber}}, {{colData}})</td>
-            </tr>
-            <tr v-for="rowNumber in Array.from('0123456789')">
-              <td v-for="colData in Array.from('ABCDEFGH')">({{rowNumber}}, {{colData}})</td>
-            </tr>
-          </tbody>
-        </table>
+        <table-view :nodeType="nodeCategory" :cardType="expandedCard" :identifier="nodeID"></table-view>
```
DoctorBud: My preference these days is to have one HTML attribute per line. The logic is that it makes collaboration easier, since changing the value of one attribute will show a Git diff only for that line and the reviewer doesn't have to figure out what changed. Also makes cutting/pasting of attrs a little easier. For future reference, don't stop the PR cuz of this.

ui/components/TableView.vue
```diff
@@ -0,0 +1,110 @@
...
+                    rows: 1000,
+                    start: start,
+                };
+                axios.get(baseURL, {
+                    params: params,
+                }).then((resp) => {
+                    _this.dataPacket = resp;
+                    resp.data.associations.forEach(function (element) {
+                        const refs = [];
+                        const sources = [];
```
DoctorBud: `sources` is defined but not used. Weird that linter didn't catch it.

ui/components/TableView.vue
```diff
@@ -0,0 +1,110 @@
...
+            onPageChange: function (evt) {
+                // { currentPage: 1, currentPerPage: 10, total: 5 }
+                this.fetchData(evt.currentPage * evt.currentPerPage);
+            },
+            fetchData(start) {
+                const _this = this;
+                const nodeType = _this.nodeType;
+                const entity_curie = _this.identifier;
+                const annotationType = _this.cardType;
+                console.log(annotationType);
```
putmantime: I forgot to remove this console.log

DoctorBud:
+1

PR #1543, opened 2018-02-22
Home Page, Better Node, TableView Error Display, USE_LOG_GOLR, more associations, close AC menu on selection

DoctorBud:
- Add new Monarch Logo graphics for navbar and home page
- Gives Node cards a specific height, rather than letting their content
  determine it.
- Adds shortcut links for debugging to About menu (e.g., Marfan, sox3).
- Add prettier, more modular Home page to SPA
- Add webpack conditional alias to enable/disable monarch-home.css
  depending on USE_SPA
- Adds USE_LOG_GOLR to enable GOLR manager debug mode.
- Generalizes the per-NodeType routes to be data-driven and DRYer.
- Remove unnecessary v-cloak from TableView.vue
- Add BioLink error display to TableView.vue
- Special case BioLink in TableView to support the /expression/anatomy
  form.
- Generally clean up the Node view, including dealing with state changes
  based on route.
- Fix jqueryUI autocomplete usage so that selection closes the menu.
- Add preliminary computeCounts() method to webapp.js to clean up
  nodeByTypeIDHandler.
- Fix issues with Patternfly/Bootstrap SCSS vars not being available in
  Vue components style section.


putmantime:
+1

PR #1544, opened 2018-02-23
filter out OMIA in autocomplete

kshefchek:
filter out OMIA in autocomplete


kshefchek:
Looks like this doesn't work because the golr manager quotes the filter string.

PR #1549, opened 2018-03-06
Update ZFIN's xrefs.json config

kevinschaper:
Hi,

I was poking around your site today and noticed that you're using the ZFIN_jump url format that we retired in 2012.  We keep it working via an apache redirect, but it's not the ideal url for our IDs.  

I also expanded the regex for matching our ZDB- ids, right now it would miss CRISPR and TALEN at least, but also a significant number of additional types of genes (lncRNA Gene, piRNA gene, etc...)

thanks!


kevinschaper:
Also, for what it's worth having the ZFIN: prefix on IDs works as well, if that simplifies anything for you. 

example: https://zfin.org/ZFIN:ZDB-GENE-990415-72

cmungall:
@kltm is this a copy of the go one?

kltm:
@cmungall It appears to be similar, but has taken a different path.
It would be very valuable to have GO, Monarch, and Alliance all on the same page with core metadata (with things like application hinting done elsewhere). At this point, a MOD wanting to actively update would have 4 locations (including local), not including and others--there should ideally be an agreed central repo for this type of broad interest metadata.

kshefchek:
+1

PR #1552, opened 2018-03-08
fix blog links

lwinfree:
change to medium links

Please check @DoctorBud :)


DoctorBud (approved):
Looks good to me. The TravisCI errors have nothing to do with your PR and should not be a blocker.

PR #1554, opened 2018-03-12
npm run dev speedups. ./install.sh required. References/Equivalent added to Node.

DoctorBud:
- Improves performance of 'npm run dev'. Use pre-minified assets. Use HotModuleReloading.
- Get used to running ./install.sh once when you download new source.
- Improves computeCounts by adding more cases.
- Adds buildXrefsString/buildXrefsList distinction, so that lists of   xrefs are available in JSON.
- Adds widgets/genExternalHrefObject to support returning of JSON Hrefs
- Moves the NodeType icon to the sidebar.
- Adds more informative Loading... message to Node
- Adds References and Equivalent Classes to Node.


putmantime:
+1

PR #1559, opened 2018-03-16
switch travis back to dev

kshefchek:
switch travis back to dev


kshefchek:
Onset classes are now a subclass of clinical course instead of clinical modifier, so we need to reintroduce a hack until https://github.com/monarch-initiative/SciGraph-docker-monarch-data/commit/00e1402c857087158b598315850d73b8199ca81d makes it through

PR #1564, opened 2018-04-02
DO NOT MERGE - BS4 Experiment

DoctorBud:
- Mostly-complete translation to BS4 for both legacy and NG Monarch 
- Removed Patternfly
- Removed use of two webpack bundles for NG


kshefchek:
+1 the tests were failing due to not having enough memory for owlsim, see https://github.com/monarch-initiative/monarch-devops/issues/114

PR #1566, opened 2018-04-05
BS4, Patternfly removal

DoctorBud:
- Mostly-complete translation to BS4 for both legacy and NG Monarch 
- Removed Patternfly
- Removed use of two webpack bundles for NG


kshefchek:
+1

putmantime:
+1

PR #1570, opened 2018-04-09
Mon autocomplete

putmantime:


ui/components/Home.vue
```diff
@@ -20,27 +20,12 @@
...
-                  id="home_search" type="text" class="form-control" placeholder="Search for...">
-              </div>
-            </form>
-
-            <p class="search-examples-text">
-              Examples: <i>Marfan Syndrome</i> <i>sox3</i>
-            </p>
+            <monarch-autocomplete
+                    homeSearch="true"
+            ></monarch-autocomplete>
```
DoctorBud: That's an unusual way to do HTML tags. Is there a logic to it? I'd move the closing tag to a separate line, but I can sort of see the value of treating the closing tag as an addendum to the closing of the opening tag. But I don't know if my editor will handle it nicely in terms of collapse/expand.

ui/components/MonarchAutocomplete.vue
```diff
@@ -0,0 +1,276 @@
...
+      return word.toLowerCase();
+    },
+  },
+  methods: {
+    debounceInput: debounce(
+      // eslint-disable-next-line
+      function () {
+        this.fetchData();
+      }, 500, { leading: false, trailing: true }),
+    fetchData(){
```
DoctorBud: Space before brace, please:
`...() {` vs `...(){`

ui/components/MonarchAutocomplete.vue
```diff
@@ -0,0 +1,276 @@
...
+    },
+    clearSearch() {
+      this.suggestions = [];
+      this.value = '';
+    },
+    categoryMap(catList) {
+      let cat1 = new Set(catList);
+      let cat2 = new Set(['Phenotype', 'gene', 'variant', 'model', 'disease']);
+      let intersection = new Set(
+        [...cat1].filter(x => cat2.has(x)));
```
DoctorBud, line 206: Given a `Set` `cat1`, is it really necessary to use `[...cat1]` to generate something that can be `filter`-ed? It seems like `Set` might be `filter`-able, so that you could write: `cat1.filter(...)`. Not blocking the PR, just asking and trying to learn here.

ui/components/MonarchAutocomplete.vue
```diff
@@ -0,0 +1,276 @@
...
+      const intArray = Array.from(intersection);
+      return intArray[0]
+    },
+    checkTaxon(taxon){
+      if(typeof taxon === 'string'){
```
DoctorBud: Space after `if` pls, so it doesn't look like a function call.

ui/components/Navbar.vue
```diff
@@ -163,37 +154,15 @@ nav#monarch-navbar.navbar.navbar-default .navbar-toggle .icon-bar {
...
-  display: inline;
-}
 
 @media (max-width: ($grid-float-breakpoint + 60)) {
   .branding-name {
     display: none;
   }
 }
 
+  .nav-ac{
```
DoctorBud, line 165: Space before brace

DoctorBud:
+1 on your PR. Please look at my comments, tho. A cursory look at the first 5 errors in the Travis log indicates that the Owlsim on beta is failing, and your code is not implicated.

ui/components/MonarchAutocomplete.vue
```diff
@@ -0,0 +1,276 @@
...
+    down() {
+      if (this.current < this.suggestions.length - 1) {
+        this.current += 1;
+      }
+    },
+    // For highlighting element
+    isActive(index) {
+      return index === this.current;
+    },
+    mouseOver(index) {
```
DoctorBud, line 183: I'm a little concerned about the necessity and correctness of this `mouseOver` and the management of `this.current`. Specifically, in `suggestionClick`, there is an `index` parameter which is passed, but `this.current` is also used in that method. Presumably, `this.current === index` when `suggestionClick` is called, and if not, something is weird. I guess the right 'fix' is to have `suggestionClick` respect its `index` parameter and do something like `this.current = index`, which would make more sense to me.

ui/components/MonarchAutocomplete.vue
putmantime: without dumping the set into an array 
```
TypeError: cat1.filter is not a function. (In 'cat1.filter(function (x) {
        return cat2.has(x);
      })', 'cat1.filter' is undefined)
```

ui/components/MonarchAutocomplete.vue
putmantime: I'm revisiting the strategy now

ui/components/MonarchAutocomplete.vue
putmantime: That said, there may be a more elegant way that I haven't considered.

ui/components/MonarchAutocomplete.vue
putmantime: I agree using `this.current` in `suggestionClick(index)` was not necessary and have changed to use `index` there.  Not sure what context you mean with `this.current = index` making more sense.   Right now `this.current` is incremented in `up() down()` and set in `mouseOver()` and `isActive(index)` returns `true` if `this.current === index` .

PR #1573, opened 2018-04-11
Add highlighting

putmantime:


ui/components/MonarchAutocomplete.vue
```diff
@@ -48,10 +51,14 @@
...
-            :title="suggestion.definition"
-            style="margin-right: 25px">
+            :title="suggestion.definition">
           <div class="row p-0">
-            <div class="col-5"><strong>{{ ...suggestion.label }}</strong></div>
+            <div class="col-5" v-if="suggestion.has_hl">
+              <span v-html="suggestion.highlight"></span>
+            </div>
+            <div class="col-5" v-else>
+              <strong>{{suggestion.label}}</strong>
```
kshefchek: We want to use match here instead of label, in our schema, label refers to the primary label (rdfs:label), where as match refers to the specific matched field in the document (primary label, synonym, curie, etc.)

putmantime: I missed that.  Changed it in both javascript and html

ui/components/Home.vue
```diff
@@ -433,7 +433,7 @@ html {
   -webkit-transition: all .3s ease-in-out;
   -moz-transition: all .3s ease-in-out;
   transition: all .3s ease-in-out;
-  border-radius: 0;
```
putmantime, line 436: commented this out for now, was overriding bootstraps soft border-radius on buttons (e.g. category filter selecter).  I didn't notice any downstream use of this, but if you know of something this is going to mess with Dan, I can punt on this

kshefchek:
+1

ui/components/Home.vue
DoctorBud: Your CSS selector is too broad; use a selector that applies this style to where you intend. Right now, any `btn` on the Home page is affected, including nested components. I question whether the above CSS is needed at all. I also think that changing the font is a bad idea. There's a default theme font in play and it makes sense to use it consistently. If your goal is simply to get the `text-transform:uppercase` style, then create a class `text-uppercase` and apply that class where needed (BS4 might already have such a utility class).

But +1, let's get this in and fix later.

putmantime: I didn't put that css selection in.  Is this a automated css configuration from the Vue cli?

putmantime: I guess not since the selection has #monarch-home-container  in it.

PR #1575, opened 2018-04-12
Fixes some broken URLs on About Monarch page

matentzn:
Fixed links:
- owlsim
- hpo
- monarch sources
- abnormality of eye
- plos PAX6 pheno image


kshefchek:
+1

PR #1578, opened 2018-04-14
fixed bug in spa version, now checks to make sure prefix is in validC…

putmantime:
…ats on mount before fetching data


DoctorBud:
+1

PR #1579, opened 2018-04-14
Production Uglify/Babel fixes. Fix search_results to not crash with bad data.

DoctorBud:
- Fixes webpack config so that a production build of NG works properly  …
with Babel and async/await.
- Improves robustness of search_results (Legacy and NG), mostly to work
  around the beta data issues with multiple categories, but also to move
more logic into JS from the HTML template.
- Eliminates use of vue-json-tree, which was there for debugging in the
  early days of MonarchNG.
- Updates babel to use babel-preset-env
- Moves the production build step from install.sh into .travis.yml, so
  that install.sh doesn't take as long and perform an unnecessary build
for ordinary development.
- Remove reference to the spa.bundle.js and spa.bundle.css that are no longer being generated or used

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,4 +1,4 @@
 {
-  "presets": ["es2015"],
-  "plugins": ["syntax-dynamic-import"]
+  "presets": ["env"],
+  "plugins": ["transform-runtime", "syntax-dynamic-import"]
 }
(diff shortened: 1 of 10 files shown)


putmantime:
+1 runs great.  I noticed autocomplete no longer facets on 'Phenotype'.  This is due to the 'Phenotype' to 'phenotype' change in the data, and will be fixed by my 'add-highlighting' PR

PR #1580, opened 2018-04-17
Monarch access search

putmantime:


ui/components/MonarchAutocomplete.vue
```diff
@@ -82,11 +82,8 @@
   </div>
 </template>
 
-<script type="text/babel">
-import axios from 'axios';
-
+<script type="text/babel">import * as MA from '../../js/MonarchAccess';
```
DoctorBud: Why is this all jammed onto one line? If there is no good reason, please keep the HTML (the script tag) and the Javascript (what's within the script tag) on separate lines.

ui/components/MonarchAutocomplete.vue
```diff
@@ -124,60 +121,33 @@ export default {
...
-        this.suggestions = [];
+      }, 500, {leading: false, trailing: true}),
+    async fetchData() {
+      const that = this;
+      try {
+        let searchResponse = await MA.getSearchTermSuggestions(this.value, this.selected);
+        searchResponse.docs.forEach(elem => {
+          const resultPacket = {
+            match: elem.match,
+            category: that.categoryMap(elem.category),
```
DoctorBud: If I understand arrow functions, it should be possible for us to use `this` instead of `that` within this `forEach`, because arrow functions don't introduce a new `this` into scope. For future reference (and pls correct me if I'm wrong).

ui/components/MonarchAutocomplete.vue
```diff
@@ -124,60 +121,33 @@ export default {
...
+      }
+      catch (e) {
+        console.log('nodeResponse ERROR', e, that);
       }
     },
     enter() {
       const currentData = this.suggestions[this.current];
       // this.$emit('value', { value: this.suggestions[this.current] });
-      this.$router.push({ path: `/${currentData.category}/${currentData.curie}` });
+      this.$router.push({ path: `/${currentData.category}/${currentData.curie }` });
```
DoctorBud: Why is there a space here? `${currentData.curie }`... Please check your PR by looking at it with GitHub's Diff Viewer, which might highlight issues your text editor isn't making obvious.

ui/components/MonarchAutocomplete.vue
```diff
@@ -204,7 +174,7 @@ export default {
     // When one of the suggestion is clicked
     suggestionClick(index) {
       const currentData = this.suggestions[index];
-      this.$router.push({ path: `/${currentData.category}/${currentData.curie}` });
+      this.$router.push({ path: `/${currentData.category}/${currentData.curie }` });
```
DoctorBud: same as above. Lose the space.

DoctorBud:
+1

Thanks for making those changes. GitHub is pretty cool in how it let's me see just the changes from your last 'style fixes' commit.

PR #1583, opened 2018-04-21
updated footer, not perfect

lwinfree:
Please check @DoctorBud. I want to change some things, but will discuss with you on Monday. Thanks!


lwinfree:
Hey @DoctorBud I made some more changes - please check it out!

lwinfree:
Alright @DoctorBud here's the latest updates. Please check them out - thanks!

ui/components/Home.vue
```diff
@@ -37,7 +37,7 @@
   </header>
 
 <!--
-  <section
+  <section>
```
putmantime, line 40: This closing '>' comes before the attributes

ui/components/Home.vue
```diff
@@ -87,16 +87,17 @@
 
   <section id="features" class="content-section text-center">
     <div class="features-section">
-      <div class="container">
-        <div class="col-lg-8 offset-lg-2">
+      <div class="container-fluid">
+        <div class="col-sm">
+          <h2>What We Do</h2>
           <ul style="list-style-type:none;">
```
putmantime, line 93: Should probably be in a class

ui/components/Home.vue
DoctorBud, line 40: This looks buggy. You are closing the tag, but there are attributes  that will not get associated. In theory, you should see bogus text on your page.

However, it looks as if this change is in a commented-out section, so it is harmless and useless.

ui/components/Home.vue
```diff
@@ -108,92 +109,107 @@
...
-          </li>
-          <li>
-            <a target="_blank" href="mailto:info@monarchinitiative.org" class="btn btn-default btn-lg"><i class="fa fa-envelope-square fa-fw"></i> <span class="network-name">List</span></a>
-          </li>
-        </ul>
+  <section class="content-section text-center">
+    <h2 class="text-center" style="margin: 10px; padding: 0px 0px 25px 0px;"> The Monarch Initiative is a collaboration between: </h2>
+    <div class="container">
+      <div class="row">
+        <div class="col-6 col-md-3">
```
putmantime, line 116: Assigning 2 column sizes to these?

ui/components/HomeFooter.vue
```diff
@@ -0,0 +1,98 @@
+<template>
+
+<div class="home-footer">
+  <div class="row">
+    <div class="col-12 col-lg-3">
+      <div class="media">
+        <img class="img-fluid"
+        style="max-height:150px; margin:auto; padding:5px;"
```
putmantime, line 8: more inline style

ui/components/Home.vue
```diff
@@ -234,7 +250,7 @@ html {
 #monarch-home-container h1,
 #monarch-home-container h2,
 #monarch-home-container h3,
-#monarch-home-container h4,
+*/ COMMENT OUT #monarch-home-container h4, */
```
DoctorBud, line 253: This looks like a mistake. If it was a proper commenting-out, the tags would be `/* comment */` instead of your `*/ comment */`.

But why are you commenting out the `h4` style only? It would be better to write more specific classes and apply them where you want, rather than globally.

DoctorBud:
+1

It's a big improvement and my comments above can be fixed in a subsequent revision.

PR #1589, opened 2018-05-02
fix behave tests

kshefchek:
Fix behave tests that I broke from changing Phenotype to phenotype in SciGraph.


DoctorBud:
+1

kshefchek:
To clarify, this PR fixes the some API and legacy functionality like autocomplete/search, autocomplete on the analyze phenotypes page, api.getPhenotypes()

PR #1590, opened 2018-05-15
webpackv4, debug msg cleanup, npm updates, nocache fix.

DoctorBud:
- Updates several NPM packages
- Migrates to webpack v4
- Removes extraneous debug messages.
- Fixes issue with the /opensearch capability so that it works on
  non-root pages.
- Updates vue-loader and adjusts the webpack config accordingly.
- Remove unnecessary HTTP header x-custom.
- Remove bogus debug messages.
- Disable nocache for GET requests. This is a stopgap measure until we can do more specific caching using ETags, unique filenames, and max-age headers.
- Fix bug where Analyze Phenotypes ortho/para-log checkboxes would crash the page if no genes were specified. This was due to a jQuery API change a while back.


kshefchek:
+1

PR #1592, opened 2018-05-22
add news, twitter, change background color

lwinfree:
OK @DoctorBud please review now :) thanks!


ui/components/HomeFooter.vue
```diff
@@ -5,7 +5,7 @@
     <div class="col-12 col-lg-3">
       <div class="media">
         <img class="img-fluid"
-        style="max-height:150px; margin:auto; padding:5px;"
+        style="max-height:125px; margin:auto; padding:5px 5px 15px 5px;"
```
DoctorBud, line 8: I think this change is responsible for the Monarch Logo being scaled incorrectly. It looks distorted (stretched horizontally). You have different padding for left-right vs up-down, which is causing the problem. The easiest solution is to have the same padding all around, or to use `margin` or a surrounding div to get the effect you want.

ui/components/Home.vue
```diff
@@ -226,6 +279,43 @@ export default {
     }
   },
 }
+
+// From https://dev.twitter.com/web/javascript/loading - 7/26/2015
```
DoctorBud, line 283: I think that the Twitter timeline looks better without some of the extra divs and classes you are using. I found this looks nice:

```html
        <div class="col-sm">
          <a class="twitter-timeline"
            href="https://twitter.com/search?q=%40MonarchInit+OR+%40hp_ontology+OR+%40GA4GH+OR+%40UDNconnect"
            data-widget-id="527567015918051328"
            xdata-theme="dark"
            xdata-tweet-limit="1"
            data-preview=""
            data-height="350"
            xdata-width="320">
            Loading Tweets about "@MonarchInit OR @hp_ontology OR @GA4GH OR @UDNconnect"
          </a>
        </div>
```

This works well when combined with adjusting the Recent Blog Posts section downward. An easy way that looks good to me is to add a `<br>`, although I'm sure there are better ways using Bootstrap's `py-2` classes. But this looked good to me when combined with the above fix:

```html
        <div class="col-sm">
          <br>
          <h4>Recent Blog Posts</h4>
```

DoctorBud:
+1 Look at my comments, please, but don't block the PR, they are easy enough to make in the next round.

lwinfree:
Thanks! Those comments all make sense to me & I will implement them in the next round.

PR #1593, opened 2018-06-01
Phenotype analysis tool

putmantime:


js/MonarchAccess/index.js
```diff
@@ -102,6 +102,56 @@ export function getNodeAssociations(nodeType, identifier, biolinkAnnotationSuffi
...
+        reject(err);
+      });
+  });
+  return returnedPromise;
+}
+
+export function comparePhenotypes(phenotypesList) {
+  const baseUrl = 'https://beta.monarchinitiative.org/analyze/phenotypes.json?';
+  const extenstion = 'input_items=';
+  const phenoCuries = [];
```
DoctorBud: I think this is better expressed using the functional paradigm:

```javascript
const phenoCuries = phenotypesList.map(elem => {
  return elem.curie;
});
```

I haven't tested the above, but it should work. If, for some reason, `phenotypesList` were not an Array, then you might need to use an Array method to invoke `map`. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map

Functional style is arguably better than imperative style. In some cases, imperative style is useful, but in the above simple transformation, it is obfuscating.

ui/components/LocalNav.vue
```diff
@@ -0,0 +1,123 @@
...
+  <div>
+    <div v-if="!dataFetched">Loading Related Terms ...</div>
+    <div v-else>
+      <div class="row border-bottom py-2">
+        <div class="col-4"><h3>Parent Terms</h3></div>
+        <div class="col-4"><h3>Equivalent Terms</h3></div>
+        <div class="col-4"><h3>Child Terms</h3></div>
+      </div>
+      <div class="row py-2">
+        <div class="col-4"
```
DoctorBud: My personal preference (but I can make a strong argument for it) when specifying an element with multiple attributes is to have ALL attributes on a separate line from the element, so that your indent level remains at 2 spaces and you don't end up with apparent inconsistency like this instance where you have one attribute `class="col-4"` on the `div` line and the other attributes oddly indented to match.

```html
<div
  class="col-4"
  v-for="parent in parents"
  :key="parent.object.id">
```

ui/components/LocalNav.vue
```diff
@@ -0,0 +1,123 @@
...
+            {{sibling.lbl}} <br/> ({{sibling.id}})
+          </div>
+        </div>
+        <div class="col-4"
+             v-for="child in children" :key="child.object.id"
+        >
+          <div @click="emitSelection(child.subject.label, child.subject.id)"
+               class="p-1 m-1 btn btn-info"
+          >
+            {{child.subject.label}} <br/> ({{child.subject.id}})
```
DoctorBud: Why are so many things stuffed into a single line? This should be:

```html
{{child.subject.label}}
<br>
{{child.subject.id}}
```

Also, `<br>` is self-closing in HTML5. Don't use `<br/>` anymore. https://developer.mozilla.org/en-US/docs/Web/HTML/Element/br

ui/components/LocalNav.vue
```diff
@@ -0,0 +1,123 @@
...
+          </div>
+        </div>
+      </div>
+    </div>
+  </div>
+</template>
+<script>
+import * as MA from '../../js/MonarchAccess';
+import * as _ from 'underscore';
```
DoctorBud: Ugh. Let's pick `lodash` or `underscore` and use just one of them. No point in having two different libs around. I recommend `lodash`, as it has the ability to surgically install only what you need.

ui/components/LocalNav.vue
```diff
@@ -0,0 +1,123 @@
...
+        this.familyData = searchResponse;
+        this.sortRelationships();
+        this.dataFetched = true;
+      }
+      catch (e) {
+        that.dataError = e;
+        console.log('BioLink Error', e);
+      }
+    },
+    emitSelection(termLabel, termId){
```
DoctorBud: SPACE BEFORE BRACE

ui/components/PhenoGrid.vue
```diff
@@ -0,0 +1,56 @@
+<template>
+    <div ref="phenogridbox"
```
DoctorBud: Same comment as above. I'd format this as:

```html
<div
  ref="phenogridbox"
  v-if="pgVersion === index"
  v-bind:id="'phenogridbox-'+index">
```

or (using your idea of a dangling `>`):

```html

```html
<div
  ref="phenogridbox"
  v-if="pgVersion === index"
  v-bind:id="'phenogridbox-'+index"
>
```

ui/components/PhenoGrid.vue
```diff
@@ -0,0 +1,56 @@
...
+          serverURL: 'https://beta.monarchinitiative.org',
+          gridSkeletonData: pgData,
+          selectedCalculation: 0,
+          selectedSort: 'Frequency',
+        });
+      },
+    },
+  };
+</script>
+<style></style>
```
DoctorBud: Split this into two lines. That way, if someone ADDs some style, we won't see the superfluous deletion of this line in the diffs.

```html
<style>
</style>
```

ui/components/PhenotypesTable.vue
```diff
@@ -0,0 +1,168 @@
+<template>
+    <div>
+       <div v-if="dataFetched">
+           <b-table :items="items"
```
DoctorBud: Alignment here would benefit from abandoning the 'first attribute goes on the element line, others go on separate lines' philosophy and instead using the 'attributes go on separate lines and not on the element line' technique.

ui/components/AnalyzePhenotypes.vue
```diff
@@ -0,0 +1,431 @@
...
+      this.phenotypes.push(payload);
+    },
+    handleGenes(payload) {
+      this.genes.push(payload);
+    },
+    generatePhenogridData() {
+      this.showPhenogrid = true;
+      this.yAxis = [];
+      this.xAxis = [];
+      this.phenotypes.forEach((elem) => {
```
DoctorBud: Try to use functional style here:

```javascript
this.yAxis = this.phenotypes.map(elem => {
  return {
    id: elem.curie,
    term: elem.match,
  };
});
```

There may even be an ES6 shorthand that obviates the `return` keyword.

js/MonarchAccess/index.js
putmantime: Much cleaner, implemented.

ui/components/LocalNav.vue
putmantime: ok, not sure where I picked up the habit of putting first attribute on same line as element

ui/components/LocalNav.vue
putmantime: good to know

ui/components/LocalNav.vue
putmantime: Fully agree, converted all to lodash surgical installations

js/MonarchAccess/index.js
```diff
@@ -102,6 +102,55 @@ export function getNodeAssociations(nodeType, identifier, biolinkAnnotationSuffi
   return returnedPromise;
 }
 
+export function getNodeLabelByCurie(curie) {
+  const baseUrl = `https://api-dev.monarchinitiative.org/api/bioentity/${curie}`;
+  const params = {
+    fetch_objects: true,
+    rows: 100,
+  };
+  const returnedPromise = new Promise((resolve, reject) => {
```
kshefchek, line 111: This chunk of code seems to be repeated several times in MonarchAccess/index.js, can we make this more DRY?

DoctorBud: Yes. That's the point of the methods in JSONAccess.js. There is currently a non-promise-based version for axios there, but a promise-based one could be easily added and might DRY up index.js.

DoctorBud:
+1

PR #1597, opened 2018-06-08
Use webpack env for configuration of external services

kshefchek:
See http://vuejs-templates.github.io/webpack/env.html, testing out using webpack to store service configurations for development and production.  This would replace the current approach of using pup tent to create global js variables.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -59,6 +59,23 @@ if (/^\w+/.test(DIST_DIR) === false || /\/$/.test(DIST_DIR) === true) { // @todo
 }
 
 log.info('webpack', `${NODE_ENV.toUpperCase()} mode`);
+if (NODE_ENV === 'development') {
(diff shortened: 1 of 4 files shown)


webpack.config.js
```diff
@@ -59,6 +59,23 @@ if (/^\w+/.test(DIST_DIR) === false || /\/$/.test(DIST_DIR) === true) { // @todo
 }
 
 log.info('webpack', `${NODE_ENV.toUpperCase()} mode`);
+if (NODE_ENV === 'development') {
+  plugins.push(
+    new webpack.DefinePlugin({
```
DoctorBud, line 64: Note that we are already using `webpack.DefinePlugin` to define other `process.env` variables. I'm guessing you've verified that the two usages will correctly setup `process.env` and not compete with each other?

PR #1607, opened 2018-06-30
add draft of new data page

lwinfree:
@DoctorBud could you please check this? The analytics page is still a placeholder, but I added a new, slim footer that can be added to other pages (besides the homepage).

Thx!


DoctorBud:
+1

Regarding the footer, I like the way that GitHub's footer is barely noticeable. Perhaps we can make the footer just text and not a background color or border? (Scroll to the bottom of this page to see what I mean).

lwinfree:
That sounds like a good idea! I'll try it out (& maybe some other ideas too) in my next PR. Thanks!

PR #1611, opened 2018-07-16
Temporarily modify conf/server_config_beta.json to point to production services

DoctorBud:
Temporarily modify conf/server_config_beta.json to point to production services


kshefchek:
Thanks! Let's also update config/dev.env.js for alpha

PR #1617, opened 2018-08-20
AssocTable styling. Home Page Search. BioLink paging.

DoctorBud:
- Add 'monarch' entry to $grid-breakpoints so navbar collapses a little later than 'md'
- Add api.js fetchAssociationCounts (plural), to retrieve counts and association numbers.
- Style literature pages in way more consistent with other pages.
- Remove recordIndex from AssocTable. Compact the table. Use chevrons for toggle.
- Implement proper paging via BioLink.
- Move list bullets in AssocTable to 'inside' from 'outside' list-style-position
- Adjust Home page so that more of the search is visible without scrolling.
- Add Search Examples to Home Page
- Move Category Filters on Home Page Search closer to search box, to reduce confusion.
- Use 'style lang="scss"' for MonarchAutocomplete.vue.

### DIFF
--- a/conf/server_config_beta.json
+++ b/conf/server_config_beta.json
@@ -7,9 +7,9 @@
 
     "scigraph_data_url" : "https://scigraph-data-dev.monarchinitiative.org/scigraph/",
 
-    "golr_url" : "https://solr-dev.monarchinitiative.org/solr/golr/",
+    "golr_url" : "https://solr.monarchinitiative.org/solr/golr/",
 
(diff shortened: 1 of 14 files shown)


kshefchek:
Is it worth reverting phenotype categories back to Phenotype.  This was an attempt to make everything consistent but I think this should have been punted on until we were very close to finishing the spa app.  There's also discussion of making all node categories upper CamelCase to be compliant with the biolink model, so we may revert back to Phenotype anyways.

https://github.com/monarch-initiative/monarch-app/pull/1589

kshefchek:
+1

PR #1624, opened 2018-09-07
BioLink for Node.vue. ui/ directory contains all of the MonarchNG code. Home Search nicer with examples.

DoctorBud:
- Move code and styles relating to MonarchNG (Next Generation) into the ui/ directory.
- Clean up the About Monarch page a little by adjusting padding/margins on the cards.
- Add a ui/data/ directory to store a relations.json and a curies.json file.
- Adjust navbar collapse point.
- Inhibit some eslint warnings.
- Delete unnecessary handleSubjectObjectInversion() kludge because BioLink now delivers subject/object in a consistent way.
- Add examples to MonarchAutocomplete.vue and restyle slightly.
- Ensure that when an autocomplete result is selected, we use the local router to change pages, rather than enduring a page reload.
- Enhance active state styling for autocomplete category buttons by lightening the border to increase contrast.
- Replace use of webapp.js and its nodeByIdHandler() with the use of a series of BioLink calls. This is primarily for Node.vue, but is implemented in MonarchAccess.


kshefchek:
Theres some redundancy with the curies.json and the [dipper curie map](https://github.com/monarch-initiative/dipper/blob/master/dipper/curie_map.yaml) to [scigraph curies function](https://scigraph-data.monarchinitiative.org/scigraph/cypher/curies) pipeline.  Is there a way to partially build this table from scigraph and then merge the rest?

DoctorBud:
@kshefchek Thanks for the feedback on those ui/data files. I removed those files and directory in my amended PR. I had been trying to use a single BL API call to get association data, and I needed the relations.json to classify the association types. But I have abandoned that effort as fruitless, and am instead making per-association-type BL calls like `/bioentity/disease/ID/genes`, so that the relation ID is hidden in the bowels of BL.

kshefchek:
+1

PR #1627, opened 2018-09-14
Support front-end working without legacy backend. Adds ui/README.md and convenience scripts. 

DoctorBud:
- Adds ui/README.md
- Adjust BS4 default link color to be more mellow
- Eliminate the no-longer-used /configuration/server.js endpoint that previously provided configuration info to the front-end SPA.
- Add 'cgrb' configuration to webapp_launcher.js to help debug CGRB migration.
- Adds mng-XXX npm scripts to make SPA (MonarchNG) easier to develop on.
- Adjust BS4 card-header padding/margin on various pages to match BS3 panel style.
- Add image assets necessary for front-end SPA
- Fix LocalNav.vue to work with BioLink
- Fix MonarchLegacy.vue to work with alternative Legacy backend, rather than relying upon webpack-dev-server's proxy.
- Add error detection to MonarchLegacy so that it gracefully reports a non-existent backend server.
- Refactor neighborhood code which was duplicated in LocalNav.vue and Node.vue. Isolate that code into MonarchAccess.
- Add transitional 'window.mngLocalServerMode=true' hack to tell SPA client to talk to alternative local legacy server. (mostly used for helping debug the front-end separation from the legacy back-end).
- Generate a 404.html file that is a copy of index.html. Mostly used for local development with http-server.


putmantime:
+1

nathandunn:
@DoctorBud  This looks great.  I added a few changes, here: https://github.com/DoctorBud/monarch-app/pull/3/files

How do I change the server-side config?   Other than that, it runs and can be meged.

DoctorBud:
Not sure what part of server-side config you are talking about? Don't expect a lot of customization options in terms of configuring the associated services; this is a transitional form that is only intended to survive the migration into a new repo, where we work more quickly and with less files. If you are talking about tweaking the BioLink/GOLR/SciGraph URLs, that is currently hardcoded as 'development' here: https://github.com/DoctorBud/monarch-app/blob/0089b95ba777d1898711ba7d4256e7f39e1a183c/ui/monarchAccess/index.js#L9

and the corresponding configuration map is here: https://github.com/DoctorBud/monarch-app/blob/mng-fe-be-separation/ui/monarchAccess/servers.js

My plan is (was?) that the bundle would be built with a default (e.g., 'production') that came from a webpack variable. We could override with a URL param (e.g., `?config=beta`), but that'd probably be unnecessary.

If (in the end) we only need a single BioLink URL to configure the front-end, then we'd add `?biolink=MYBIOLINKSERVER`.

I'm making the conservative assumption that our BL servers will NOT be colocated with our UI servers, so I didn't rely on tricks assuming nginx or relative paths. In theory, we should be able to put the UI onto GitHub pages and all BL servers should support CORS.

Issue #1591, opened 2018-05-17
GDR mismatch


vgainullin:
This GDR appears to be a bad mapping
https://monarchinitiative.org/gene/HGNC%3A27172#diseases

How is the following link established? 
GAREM2 -> mitochondrial trifunctional protein deficiency | pathogenic_for_condition

kshefchek:
This is inferred from one or more variant to disease associations, here is one: https://www.ncbi.nlm.nih.gov/clinvar/variation/194917/

My understanding is this snp occurs in both HADHA and GAREM2, as they overlap on opposite strands.  I believe this snp is more commonly associated with HADHA because it is located at a splice site, whereas it is an intronic variation in GAREM2.

I don't believe this is an error in our ingest, but rather we are not showing evidence adequately for a user to make this distinction, and we should also weight the association towards HADHA.

vgainullin:
There is currently no disease in association with GAREM2, so this listing is misleading.

kshefchek:
@vgainullin yes you are right, if there is not a gene to disease association we should not show one.  See https://github.com/monarch-initiative/dipper/issues/593 for the data ingest ticket, will use this ticket for addressing UI issues.  We likely still want to show that GAREM2 contains an intronic variant that is linked to this disease, but not make the gene to disease association directly since the gene is likely not affected by the variation. See GeneCards for GAREM2 as an example: https://www.genecards.org/cgi-bin/carddisp.pl?gene=GAREM2

kshefchek:
We've removed the incorrect inference, see https://monarchinitiative.org/gene/HGNC:27172 , @vgainullin thanks for reporting!

PR #1637, opened 2018-11-30
Add filter for sources for gene to disease tables

kshefchek:
This PR adds a filter on gene to disease views that requires at least one of a set of sources, currently omim, orphanet, gwascatalog, omia.  This will remove gene to disease associations where either ctd, clinvar, or coriell are the only sources.  See as a background:

https://github.com/monarch-initiative/monarch-app/issues/1591
https://github.com/monarch-initiative/dipper/issues/593
https://github.com/monarch-initiative/monarch-cypher-queries/issues/23
https://github.com/monarch-initiative/dipper/pull/602
https://github.com/monarch-initiative/dipper/issues/685

### DIFF
--- a/lib/monarch/web/webapp.js
+++ b/lib/monarch/web/webapp.js
@@ -3361,9 +3361,23 @@ function diseaseByIdHandler(request, id, fmt) {
     addGolrTable(info, 'subject_closure', 'phenotypes-table', phenotype_filter, 'disease_phenotype','#phenotypes');
     info.phenotypeNum = engine.fetchAssociationCount(phenotype_filter, 'object');
 
+    const required_sources = [
+        'https://data.monarchinitiative.org/ttl/omim.ttl',
+        'https://data.monarchinitiative.org/ttl/orphanet.ttl',
+        'https://data.monarchinitiative.org/ttl/omia.ttl',
+        'https://data.monarchinitiative.org/ttl/gwascatalog.ttl'
+    ];
+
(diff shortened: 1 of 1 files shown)


kshefchek:
@pnrobinson @cmungall et al please advice on this source list, should we include Orphanet or only associations with Orphanet + OMIM

cmungall:
I believe Peter's recommendation is that we do include Orphanet

pnrobinson:
I would like to see us explore the Orphanet scheme for relating genes to phenotypes (seven categories). This would give us a better way to present modifier genes and other classes other than direct causative. I think this will need a little thought and exploration, I am not familiar with this resource yet. 
I imagine that OMIM can be our primary source that we can just trust. I would say we should explore adding additional sources for individual variants with human biocuration, and once we think we understand the issues such as the Orphanet file we can expand with testing. 
There are probably few if any valid Mendelian G2D associations that are not in OMIM.
A further recommendation is to indicate the level of confidence according to ClinGen. 
Let's talk about this!

kshefchek:
Agreed.  Changing these mappings is now a matter of adjusting some lines in our yaml configs for orphanet (thanks to @TomConlin), https://github.com/monarch-initiative/dipper/blob/master/translationtable/orphanet.yaml

kshefchek:
see also: https://github.com/monarch-initiative/dipper/issues/431
https://github.com/oborel/obo-relations/issues/146

historical: https://github.com/monarch-initiative/dipper/issues/67

